PR #5, opened 2021-04-02
Change video link to download link

helena-intel:
With this change, clicking on the link near the monodepth result video will download the video, instead of Jupyter trying to display it, resulting in an error. I tested only on Windows, but adding "download" to the link is a standard. See https://caniuse.com/?search=download%20attribute


ryanloney:
LGTM

ryanloney (commented):
LGTM

PR #8, opened 2021-04-03
Updates to README, added jupyterlab gif and fixed video download URL

ryanloney:
This includes the following changes from me:
- Added notebooks/jupyterlab.gif (gif that Raymond created)
- Embedded jupyterlab.gif in README after Step 6
- Fixed the Discord URL
- Fixed a typo in note about Command Prompt before Step 1
- Instructions to remove env from jupyter kernel (already in main)

This also includes Helena's PR: 
- Change video link to download link #5

Please review and merge.


README.md
```diff
@@ -86,14 +90,54 @@ jupyter lab
 
 In Jupyter Lab, select a notebook from the file browser using the left sidebar. Each notebook is located in a subdirectory within the `notebooks` directory.
 
-## Troubleshooting
+<img src="notebooks/jupyterlab.gif">
```
helena-intel, line 93: Should we say `jupyter lab notebooks` instead of `jupyter lab` so that the notebooks directory is the root directory? Currently the user always has to navigate to the notebooks directory first, which is an extra step, and potentially a bit confusing.

For a gif in our main README, it is nicer if there is only an openvino_env kernel and not also a venv_openvino kernel. I would also strongly suggest making it an mp4 instead of a gif, because of accessibility issues with animated gifs. In a README you cannot just look away from the gif if you are bothered by it- you need to read the instructions around it. It is also nicer when it is a video that you explicitly click because then it is clearer where the start is. I am not sure if we need this gif or video here in the main README actually. To me, these video-instructions are nice in a (to be added) "hello-jupyter" README.  @ryanloney @raymondlo84

raymondlo84 (approved):
=) Looks like the download video is fixed! =) Awesome changes.

README.md
ryanloney: I updated the `jupyter lab` part to `jupyter lab notebooks` I agree we should fix the GIF as you suggest, but let's do that later. If either of you have experience making the MP4 and embedding in markdown, please make a PR for it.

PR #13, opened 2021-04-07
updated troubleshooting on readme

ryanloney:
I made the following changes to README.md: 

- Added note in troubleshooting about installing OpenGL on Fedora, Red Hat and Amazon Linux 2 in order for OpenCV to work
- Added note in troubleshooting that links to Rosetta community discussion for macOS developers with M1


helena-intel:
Thanks Ryan! I think we should make a proper system requirements page that lists the required system dependencies (like libpython3.7 and Mesa). But for now it is good that this is here. I also saw the missing libpython3.7.so.1 error on a RedHat cloud server. And now that Ubuntu 20 and Python 3.8 is supported it can also be libpython3.x.so that is missing, where x is [6,7,8].

helena-intel (approved):

raymondlo84:
Are we still against the idea of the one line sudo? 

sudo apt-get install _____________ with all the packages for each Linus environment?

raymondlo84 (approved):
The trouble shooting is getting very long, and maybe we should consider putting them in a nicer way soon.

ryanloney:
> The trouble shooting is getting very long, and maybe we should consider putting them in a nicer way soon.

Good idea. Let's talk about it tomorrow

ryanloney:
> Thanks Ryan! I think we should make a proper system requirements page that lists the required system dependencies (like libpython3.7 and Mesa). But for now it is good that this is here. I also saw the missing libpython3.7.so.1 error on a RedHat cloud server. And now that Ubuntu 20 and Python 3.8 is supported it can also be libpython3.x.so that is missing, where x is [6,7,8].

I've only seen this issue when using deadsnakes to install older versions of Python on Ubuntu, so not sure if it matters for Python 3.8. Let's talk about it tomorrow.

PR #14, opened 2021-04-07
added Python 3.8 to support matrix for Linux

ryanloney:
I made the following changes to README.md 
- Added Python 3.8 only to Linux distributions on the support matrix
- Removed Python 3.7 from RHEL (since 3.6 and 3.8 are only officially supported on RHEL) 
- Added discussions URL to NOTE at the top of README
- Changed Getting Started to include 3.8 in Python range 
- Removed NOTE about Python 3.8 not being supported yet from the System Requirements 

I did not add Python 3.8 for Windows or macOS yet. Please change if you see fit.


helena-intel (approved):
🎉

raymondlo84 (approved):

PR #15, opened 2021-04-07
102-pytorch-onnx: resize segmentation mask, add Coco image

helena-intel:
Segmentation mask is resized to original image size for side-by-side showing of image and result. Image changed to Coco on crosswalk.
image


raymondlo84 (approved):
We can go with a compressed PNG for this. I think we should save some space if possible.

ryanloney (approved):
LGTM

PR #16, opened 2021-04-07
Add Python 3.8 to nbval.yml and limit nbval runs

helena-intel:
Do not run notebook validation when only markdown files are changed.

### DIFF
--- a/.github/workflows/nbval.yml
+++ b/.github/workflows/nbval.yml
@@ -1,6 +1,12 @@
 name: nbval
 
-on: [push, pull_request]
+on: 
+  push:
+    paths-ignore:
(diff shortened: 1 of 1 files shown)


raymondlo84 (commented):
Um... Helena how do we validate this? I mean is that part of the CI?

ryanloney:
> Do not run notebook validation when only markdown files are changed.

Thank you. I realize I am guilty of triggering many CI jobs when the .md files get updated.

helena-intel:
> Um... Helena how do we validate this? I mean is that part of the CI?

Yes, these files are for the CI, if they are invalid, the CI will immediately fail.

PR #17, opened 2021-04-08
Add quotes around paths in MO command and remove deprecated option from MO

helena-intel:
Add quotes around sys.executable and output_dir to allow paths with spaces.
Remove `--move_to_preprocess` because that option is deprecated. Specifying means and scale moves these values to preprocess automatically (tested with ONNX notebook that output still looks the same)


raymondlo84 (approved):
Should we put that in main? instead of develop for this type of fixes?

ryanloney (approved):

PR #18, opened 2021-04-08
Update main with changes from develop

ryanloney:


helena-intel:
Failing CI checks are memory errors for PyTorch notebook. It only fails on Windows, and not all the time. I made a PR to develop that should fix this for future CI tests.

helena-intel (approved):

PR #23, opened 2021-04-13
Improve check_install script

helena-intel:
- Add Python 3.8 to supported Python versions
- Improve check for whether OpenVINO is installed globally
- Add troubleshooting when OpenVINO is installed globally
- Add check for whether OpenVINO (non-dev) is installed
- Add check that tests whether importing fails (catching missing libpython-dev)

We should in the future have pages that explain some of these steps in a bit more detail (like how to fix the "global OpenVINO installed in your PATH/LD_LIBRARY_PATH" problem). But for now this at least gives some pointers.

Tested with a few scenarios on Windows and Linux (Ubuntu 20.10). If anyone can test on macOS, that would be good!

### DIFF
--- a/check_install.py
+++ b/check_install.py
@@ -1,13 +1,32 @@
+import os
 import sys
 from pprint import pprint
 
 PYTHON_EXECUTABLE = sys.executable
 PYTHON_VERSION = sys.version_info
+OS = sys.platform
 
 IN_OPENVINO_ENV = "openvino_env" in sys.executable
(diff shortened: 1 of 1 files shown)


raymondlo84 (commented):
Passed my Mac OSX.
(openvino_env) rlo@rlo-mac01 openvino_notebooks % python check_install.py 
System information:
Python executable: /Users/rlo/Documents/openvino_dev/official_notebook/openvino_notebooks/openvino_env/bin/python
OpenVINO environment activated: OK
Python version: 3.7 OK
OpenVINO development tools installed: OK
Numpy version: 1.18.5 OK

Everything looks good!

ryanloney (approved):

PR #48, opened 2021-04-29
Change instructions for pip install to upgrade pip

helena-intel:
Use `--use-deprecated=legacy-resolver` with latest pip version to handle dependency conflicts instead of downgrading pip.
The danger is that pip may delete this legacy-resolver option. But hopefully we will not need this option anymore when that happens. If it does happen, we can always instruct to downgrade pip again to a version that still has this option. 
The benefit is that it is good practice to upgrade pip and this ensures that installing the requirements works when software packages are updated. This also makes the notebook work on macOS X Big Sur.

Also added new method to CI, so if it does fail at some point, the CI will catch it.

### DIFF
--- a/.github/workflows/nbval.yml
+++ b/.github/workflows/nbval.yml
@@ -23,8 +23,8 @@ jobs:
         python-version: ${{ matrix.python }}
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip==20.2.*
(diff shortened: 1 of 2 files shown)


raymondlo84 (approved):
It seems to work! Tested on Mac OSX 10.15...

ryanloney (approved):
Works for me too!

PR #50, opened 2021-04-29
Superresolution notebook

helena-intel:
Separate video and image notebook because the image method is somewhat large and complicated. 

Use high quality images for good results. Model also works well to upsample downsampled images. Recommended settings: 1032 model with a CROP_FACTOR of 2, for a 2x superresolution output. See notebook for model source.

Still TODO:
- Add README and intro text
- Add some more explanation, especially about the crop methods.
- Test CROP_FACTOR with 3x model
- Use Coco input image - not too large though, and one that works well.


ryanloney (approved):
Looks good. Once it is in develop branch I will add some descriptions about the model and notebook.

PR #51, opened 2021-04-30
Add options to speed up monodepth video creation

helena-intel:
- option to reduce FPS of monodepth video
- option to reduce frame width/height of monodepth video
- style improvements (constants in uppercase, variables in lowercase, all settings in settings cell)

Note: I am not a fan of this huge video inferencing loop at all. I want to refactor that for all notebooks. But since it is already such a large loop, these options do not increase the complexity that much and the speedup is large.


ryanloney (approved):
This whole notebook runs in 20 seconds on my laptop now. Wow! I suggest we switch back to Coco on the bike, since we use Hollywood in the U2^Net notebook (and save a tiny bit of space by deleting one of the images 😊)

helena-intel:
Hollywood Coco is so great :) I'll switch it back in a new PR, will merge this ti develop now.

PR #52, opened 2021-04-30
Add safety gear detection notebook

helena-intel:
This notebook is a bit simple as is, and the safety gear model has issues, but there is interest in safety gear demos with OpenVINO, and I think the notebook is useful enough as is to have in the develop branch. 

Notebook uses model and video from https://github.com/intel-iot-devkit/safety-gear-detector-python
I converted the model to IR and use the IR model directly here. The source model is Caffe and installing Caffe is not trivial, so not something we want to include in our requirements (for now at least).


ryanloney:
Screen Shot 2021-05-01 at 09 25 40
I'm fine merging into develop, but we may want to clarify what type of video is best for this model. It clearly does not perform well from this overhead view where it is identifying caution tape on the floor as safety gear. Do we have any other sample videos that are from the same vantage point as the videos provided by the original video? I will try to find some to test.

ryanloney (approved):
Approve, but need to test further and understand limitations before putting this into main branch.

raymondlo84:
When it thinks the floor is a hat, we definitely should not release it yet.

helena-intel:
Cleaning up PR's, merging to develop-staging.

PR #54, opened 2021-04-30
Change Tensorflow notebook to train for 5 instead of 15 epochs

helena-intel:
This decreases the time it takes to run this notebook (also in the CI).
Added message that user can increase the number of epoch for better results.
Result on demo image is still correct.

### DIFF
--- a/notebooks/103-tensorflow-training-openvino/103-tensorflow-training-openvino.ipynb
+++ b/notebooks/103-tensorflow-training-openvino/103-tensorflow-training-openvino.ipynb
@@ -849,7 +849,10 @@
    },
    "outputs": [],
    "source": [
-    "epochs = 15\n",
+    "# Increase the number of epochs to train the model for longer.\n",
+    "# Training for more epochs will likely improve the accuracy of the model.\n",
+    "# In the original TensorFlow notebook, the model is trained for 15 epochs.\n",
+    "epochs = 5\n",
     "history = model.fit(\n",
     "  train_ds,\n",
     "  validation_data=val_ds,\n",
@@ -1147,7 +1150,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.10"
+   "version": "3.6.8"
   }
  },
  "nbformat": 4,


raymondlo84:
I would go with at least 10 to show the training is getting less of a diminishing return. 

Also, you are right we need to install 'curl' or the last part will fail with Ubuntu.

ryanloney:
> I would go with maybe 7-8 to show the training is getting less of a diminishing return.
> 
> Also, you are right we need to install 'curl' or the last part will fail with Ubuntu. I saw the error just now on Ubuntu.

Can we change this to not use curl? 
```
output = subprocess.check_output(f"curl {inp_img_url} --output {inp_file_name}", shell=True)
```

ryanloney (approved):

raymondlo84:
output = subprocess.check_output(f"curl {inp_img_url} --output {inp_file_name}", shell=True)
 

Yes, what's the alternative that you love the most?

ryanloney:
> output = subprocess.check_output(f"curl {inp_img_url} --output {inp_file_name}", shell=True)
> 
> Yes, what's the alternative that you love the most?

Helena used urllib in monodepth notebook. Can we use that? https://docs.python.org/3/library/urllib.html

helena-intel:
> Helena used urllib in monodepth notebook. Can we use that? https://docs.python.org/3/library/urllib.html

I'll make some standard functions that we can use across notebooks this week. This notebook is only in the develop branch now, so this is not urgent to fix in my opinion.

helena-intel:
I changed it to five epochs because this is a 100-series notebook that should run relatively quickly on all computers (and the CI...). But we discussed to move this notebook to a different series, for which we can run different tests. With that change, I think it is better to keep the number of epochs at 15 and have the TensorFlow part exactly the same as the TensorFlow notebook. We can than also uncomment the early training part (that is part of the tutorial flow).

raymondlo84:
Yes, let's move these out of the '1 series' I wonder how level 40 would look like now... Are we going to launch a rocket to Mars with AI?

helena-intel:
Closing this PR in favor of keeping the number of epochs and changing the name. In a future PR we will not test this complete notebook in the CI.

PR #55, opened 2021-04-30
Add note to only use crop factor with 1032 model

helena-intel:
No code changed, only added note.


raymondlo84:
Looks good. What about video one? Should we do the same approach?

raymondlo84 (approved):

ryanloney (approved):

helena-intel:
> Looks good. What about video one? Should we do the same approach?

Video is a bit different, but yes, we can do the 2x video superresolution by having an input video that is twice the input size and then downsampling to input size.

PR #60, opened 2021-05-02
Create README.md for U^2-Net demo notebook

helena-intel:
Basic README. We should expand this, but at the moment there is no README for this notebook and we should at least give some indication what the notebook does.

### DIFF
--- a/notebooks/205-vision-background-removal/README.md
+++ b/notebooks/205-vision-background-removal/README.md
@@ -1,0 +1,24 @@
+# Background Removal Demo
+
+This demo notebook shows image segmentation and removing/adding background with [U^2-Net](https://github.com/xuebinqin/U-2-Net) and OpenVINO.
+
+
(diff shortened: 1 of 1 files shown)


helena-intel:
According to our settings all CI checks need to pass, but we disabled this for markdown-only commits, so the CI will not run for this commit.

raymondlo84:
That's a great addition. I still want to see some sort of 'gallery' mode at the front page so we can discover what's new or what's inside these notebooks from the front page.

ryanloney:
> That's a great addition. I still want to see some sort of 'gallery' mode at the front page so we can discover what's new or what's inside these notebooks from the front page.

Do you have any suggestions on how to implement this idea? Make a GIF that is updated manually? Or some more scalable approach?

ryanloney (approved):
Looks good. Thank you for adding it. Where are the images hosted?

helena-intel:
> Where are the images hosted?
I pasted them from the clipboard and github made these githubusercontent links. I kind of like it for these non-essential images, but I could also add it them to the repo.

PR #61, opened 2021-05-03
Added descriptions to superres notebooks

ryanloney:
Added descriptions with citation of the research paper. Also added NOTE about using raw images and not using compressed images like JPEG for the image notebook and NOTE about the SISR model not being ideal for video (with ask to suggest MISR models we should try).


helena-intel:
Thanks Ryan! Could you also put this description in a README.md? It would be nice to add a bit more there, but since there is no README now it is nice to at least show something.

Let's see if people respond to your ask for suggestions! I would love to add a state of the art video resolution model. I do have a DCSCN notebook in development that I would like to try with an open source model. I'm also adding the caveat (for background) that superresolution tends to be not the kind of model where you just replace the path to the model file. But the complexity is also what makes it good for a demo.

helena-intel (approved):

PR #62, opened 2021-05-03
Add .gitignore

helena-intel:
Standard Github Python .gitignore, minus a few things that we definitely don't need (like Django), plus openvino_env

It seems harmless to leave most of it here, like the spyder lines, in case someone does use that.


ryanloney (approved):
Awesome. This is very comprehensive.

PR #74, opened 2021-05-13
max_retries option for video downloading in superresolution notebook

helena-intel:
YouTube downloading sometimes fails in CI. It can also fail when running a notebook manually. This PR adds a `max_retries=2` option to the YouTube download code, to make the download code a bit more robust. 

This max_retries option was added in PyTube 10.7, so I added that to the requirements. 10.7 was released beginning of April, so this will not affect existing notebook users who followed our instructions, since we added the superresolution notebook after that. If you have a not-up-to-date virtualenv, you should update PyTube.


helena-intel:
max_retries does not solve this PyTube download issue, that seems to be consistent on Python 3.8 and Ubuntu 18. I'll look into it.

helena-intel:
Closing in favor of https://github.com/openvinotoolkit/openvino_notebooks/pull/84

Issue #77, opened 2021-05-17
Python 3.7 from Windows Store = DLL Import Error


raymondlo84:
https://www.microsoft.com/en-us/p/python-37/9nj46sx7x90p?activetab=pivot:regionofsystemrequirementstab

I got a DLL import error if I use Python 3.7 version from the Windows Store. I believe this is something others should be aware of?  

-ray

helena-intel:
That is annoying indeed. It is strange that Microsoft Store doesn't just install the regular Python. I added a note to the wiki. https://github.com/openvinotoolkit/openvino_notebooks/wiki/Windows  
We now link to the wiki in the README (currently in develop) so hopefully people who run into this will find that.

raymondlo84:
The instruction is updated (README.md) and we should use the official Python 3.7 for now.

PR #87, opened 2021-05-22
Add note about Microsoft Store version of Python to README

helena-intel:
Add "The Microsoft Store version of Python is not recommended." to README, before link to wiki with Windows installation instructions.

See https://github.com/openvinotoolkit/openvino_notebooks/issues/77

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ or create an [issue](https://github.com/openvinotoolkit/openvino_notebooks/issue
 ### Windows
 
 - On Windows, if you have installed multiple versions of Python, use `py -3.7` when creating your virtual environment to specify a supported version (in this case 3.7).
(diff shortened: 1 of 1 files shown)


ryanloney (approved):

raymondlo84 (approved):
I think it does not work with the DLL error. Do we have a workaround for that?

helena-intel:
> I think it does not work with the DLL error. Do we have a workaround for that?

IMO the workaround is: "use an installer version of Python". See also https://dev.to/naruaika/why-i-didn-t-install-python-from-the-microsoft-store-5cbd

helena-intel:
Closing in favor of https://github.com/openvinotoolkit/openvino_notebooks/pull/92

PR #90, opened 2021-05-27
s2i branch

helena-intel:
There are currently two Dockerfiles: Dockerfile pulls the notebooks from the git repository (this is what we used so far). Dockerfile.local copies the files directly. That seems better.
I also added the energy notebooks to "testing_notebooks". I can remove them if you want. I find it useful for testing, and we will add these notebooks to the repository.
There are some pip installs in the Dockerfile that should be moved to the Pipfile lock for the final release.


ryanloney (approved):
Thank you for creating this, looks good to me.

PR #93, opened 2021-05-28
Catch YouTube timeout errors in superresolution notebook

helena-intel:
Continuing to make YouTube downloading more robust. Catches TimeoutError and socket.timeout in addition to HttpError. Local video file is used instead of raising an exception.


raymondlo84 (approved):
Looks clean to me.

PR #110, opened 2021-06-04
Use older jedi to fix the autocomplete issue

tomdol:
A lot of `ipython` releases are affected by an incompatibility with the newest release of `jedi`. This causes exceptions when pressing tab in a notebook to use autocompletion and/or code suggestions. The issue is described here https://github.com/ipython/ipython/issues/12740 and the easiest fix is to downgrade `jedi` to 0.17

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,5 +26,6 @@ jupyterlab
 # Pin versions to prevent known dependency issues
 importlib-metadata==3.7.3
 ipython==7.10.*
+jedi==0.17.2
 numpy==1.18.*
 setuptools>=56.0.0


helena-intel (approved):
Thanks for this PR! I do not generally like to pin to older versions unless there is a good reason, but autocomplete is a really nice to have feature indeed.  I had expected this would have been fixed differently by now. I am fine with this, assuming there are no security issues with jedi==0.17.2.  

Another possible solution is https://github.com/krassowski/jupyterlab-lsp . That adds smarter autocomplete and other nice-to-have features, but it is not something to include in the standard requirements.

ryanloney (approved):
Thanks @tomdol for catching this issue

raymondlo84:
So it will fix the tab vs space paradox? Should we go with space or tab in our notebooks by the way =)?

PR #115, opened 2021-06-11
Add s2i build files to main branch

ryanloney:
I've added a Dockerfile and two directories: .s2i and .openshift. These contain the files needed to build the OpenShift image for the OpenVINO Operator.


Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+# Copying in override assemble/run scripts
+COPY .s2i/bin /tmp/scripts
+# Copying in source code
+COPY .openshift /tmp/src
+# Change file ownership to the assemble user. Builder image must support chown command.
+RUN chown -R 1001:0 /tmp/scripts /tmp/src
+USER 1001
+RUN /tmp/scripts/assemble
+
+# These manual pip installs will be removed before final release and added to the Piplock file
```
helena-intel, line 39:
```suggestion
```

Suggest to remove the line that we will remove this :-) We should be able to remove this in 2021.4, but need to verify. No need to add this to the source IMO.

helena-intel (changes requested):
Thanks Ryan, this looks good. I have one minor suggested change.

Dockerfile
helena-intel: Merging this now, since this has no impact on functionality. Can change later.

helena-intel:
I merged this now. There are a few things to do/investigate for a future PR:
- building on Windows doesn't work, with Git line-end conversion settings. We should look into this.
- because of dependency conflicts, openvino-dev is installed separately in the Dockerfile. This should be done with pipenv after  2021.4 release.
- this adds a 17MB binary to the main repo, which is not ideal. If it cannot be removed, we should consider moving these files to a separate branch and keep that in sync with the main branch
- The s2i build should be added to the CI.

PR #121, opened 2021-06-14
Test requirements install

helena-intel:
Add a test to check that installation of requirements works. This is part of the regular nbval test, but useful if you just want to check manually whether there are problems with installation.
Also added separate check to see that installation with the mirror in the instructions for China still works. Not a perfect test, since we will not be testing from within China, but better than nothing. This test runs daily.


ryanloney (approved):
Thanks for adding the China mirror.

PR #123, opened 2021-06-17
Remove unnecessary s2i openshift file

ryanloney:
This file was carried over by accident when testing the build process. It is not actually needed and should not be here.


helena-intel (approved):
Thanks Ryan! Happy with the reclaimed space.

PR #124, opened 2021-06-18
Updated dependencies and base layer for openshift container

ryanloney:
Updated some of the dependencies used in Pipfile.lock, requirements.in and Dockerfile.


Dockerfile
```diff
@@ -37,7 +37,7 @@ USER 1001
 RUN /tmp/scripts/assemble
 
 # These manual pip installs will be removed before final release and added to the Piplock file
-RUN pip install openvino-dev
+RUN pip install openvino-dev tensorflow~=2.4.1 tensorflow-serving-api --use-deprecated=legacy-resolver
```
helena-intel, line 40: Why did you move the tensorflow installs from the pipfile to here?

.openshift/Pipfile
```diff
@@ -15,12 +15,10 @@ geffnet = "==0.9.8"
 fastseg = "*"
 onnx = "*"
 ipywidgets = "*"
-torch = "==1.5.1"
-torchvision = "==0.6.1"
+torch = "<=1.7.1+cpu,>=1.5.1+cpu"
```
helena-intel, line 18: Should we pin PyTorch to a specific version here, since it will be pinned to a specific version in the Pipfile.lock file anyway?

helena-intel (commented):
Thanks Ryan! I am happy to see Jupyter Lab 3! Did you test that everything works?

.openshift/Pipfile
ryanloney: Good idea. Updating it

Dockerfile
ryanloney: pipenv was not able to do the dependency resolution. once we have 2021.4 this will be easier

.openshift/Pipfile
ryanloney: I could not get this to work by changing the Pipfile and requirements.in. Let's leave it like this for now. I want to address the security issues first.

ryanloney:
Ready to merge if you are ok with us revisiting the PyTorch part later @helena-intel? I will need to update this for 2021.4 release anyway.

helena-intel:
> Ready to merge if you are ok with us revisiting the PyTorch part later @helena-intel? I will need to update this for 2021.4 release anyway.

Yes, the Pytorch version is not important, thanks, I'll merge this.

helena-intel (approved):

PR #125, opened 2021-06-18
Add installation and launch script, with tests

helena-intel:
install_notebooks.py installs the notebook in a virtualenv in the current directory. launch_notebooks.py launches the notebooks from the virtualenv in the current directory or the parent directory. 

Added a test to verify that the install works. Did not add anything about the install_notebooks.py script to the README yet - not sure if we want to mention this as a recommended method already.

### DIFF
--- a/.ci/utils.py
+++ b/.ci/utils.py
@@ -1,0 +1,26 @@
+def check_jupyter_running():
+    """
+    Check that Jupyter Lab is running. Raises ValueError if not. This is a utility function 
+    for the CI. It is not suitable for use outside of the CI.
+    """
+    import psutil
+    for item in psutil.process_iter():
+        if "python" in item.name():
+            print(item.cmdline())
+            if len([line for line in item.cmdline() if "jupyterlab" in line]) > 0:
+                # Jupyter found in list of running processes: return without error
+                return
+    raise RuntimeError("Jupyter Lab is not running")
+
+def check_install_output():
+    """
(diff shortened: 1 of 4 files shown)


ryanloney:
I tested on a macbook and Ubuntu 20.04 machine. My Ubuntu machine is headless, so when I run launch notebooks, it does not allow traffic aside from localhost. If someone is trying to run notebooks on a NUC or other device without a screen, this may be something that they want to configure. Is there an easy way to have a launch option that will add `--ip=0.0.0.0` to the jupyter command? 

It did work on both my mac and ubuntu machine. Even typing `python` and not `python3` it still worked. Very nice!

ryanloney (approved):

helena-intel:
> I tested on a macbook and Ubuntu 20.04 machine. My Ubuntu machine is headless, so when I run launch notebooks, it does not allow traffic aside from localhost. If someone is trying to run notebooks on a NUC or other device without a screen, this may be something that they want to configure. Is there an easy way to have a launch option that will add `--ip=0.0.0.0` to the jupyter command?

 `python launch_notebooks.py --ip=0.0.0.0 --no-browser` should already work! You can add all the flags that you can add to `jupyter lab`.

helena-intel:
> It did work on both my mac and ubuntu machine. Even typing `python` and not `python3` it still worked. Very nice!

That was by design but I see that it doesn't work in all cases anymore - I will look into that. I do want that to work.

PR #127, opened 2021-06-19
Patch for aiohttp cve

ryanloney:
Getting pipenv to work has been very difficult. Adding this patch for aiohttp cve (using pip install in Dockerfile) until we have a new Pipfile for 2021.4.


ryanloney:
I was able to get the pipfile.lock updated with aiohttp and tensorflow 2.3.3. Updated with a new commit.

helena-intel:
> I was able to get the pipfile.lock updated with aiohttp and tensorflow 2.3.3. Updated with a new commit.

Thanks Ryan, that is much nicer!

helena-intel (approved):

PR #129, opened 2021-06-20
Add notebooks test for Docker image

helena-intel:
Tests that the notebooks run in the Docker image.

I installed nbval with pip install in the Dockerfile for now. Will do that with pipenv after 2021.4 release. 

@ryanloney the Dockerfile now installs tensorflow 2.5 (because pip install tensorflow-estimator updates the tensorflow installed by pipenv). Notebook tests pass.


ryanloney (approved):

raymondlo84:
Should we enable the onednn flag? that will 2x the training speed?

helena-intel:
> Should we enable the onednn flag? that will 2x the training speed?

If we install TensorFlow 2.5, we should. But then we should include TF 2.5 in the requirements.in/Pipfile.lock. We can also do that after 2021.4 release next week.

helena-intel:
> Should we enable the onednn flag? that will 2x the training speed?

If we install TensorFlow 2.5, we should. But then we should include TF 2.5 in the requirements.in/Pipfile.lock. We can also do that after 2021.4 release next week.

helena-intel:
I added 2021.4 wheels. Will remove that before merging once 2021.4 is on PyPI. Will also update pipenv then, and use TF 2.5.

raymondlo84:
What are our official commands to compile and test this docker file?

helena-intel:
> What are our official commands to compile and test this docker file?

You build it with `docker build -t openvino_notebooks .` in the directory that contains the Dockerfile and then run `docker run -it --entrypoint /tmp/scripts/test openvino_notebooks`

helena-intel:
I cancelled the nbval tests because they are not relevant for this PR. docker_nbval tests pass.

PR #130, opened 2021-06-22
Add PaddleGAN AnimeGAN notebook

helena-intel:
AnimeGAN notebook with model from https://github.com/PaddlePaddle/PaddleGAN

Convert PaddleGAN model to ONNX and then to IR, and show inference results.

PaddlePaddle requirements are installed in the notebook, with `!pip`. This requires that users activated the openvino_env environment and kernel - which they do if they follow our instructions. 

Converting this model was not completely straightforward. I added some steps to the notebook that show how to go about this, for example to do `predictor.run??` to show the source of the function, to see how to preprocess and postprocess the model output. 

image

This is a Draft PR - a README should be added and the descriptions in the notebook should be updated before merging.

The notebook currently fails in the CI for Windows. I'll look into that - it seems to be a resource issue. It works on my Windows laptop.


raymondlo84:
There are some random extra lines and spaces around the code. Worth a while to do a quick clean up. But the code works well on my machine, and my biggest question is the final comparison. We should maybe run 2-3 images and do an average? I don't know why the PaddlePaddle is so slow by the way... Is there overhead and also a fair comparison? 

e.g., for some of the inconsistent new lines and spaces... 

# Step 1. Load the image and convert to RGB

image_path = Path("coco_bike.jpg")

raymondlo84:
[06/22 20:37:08] ppgan INFO: Downloading animeganv2_hayao.pdparams from https://paddlegan.bj.bcebos.com/models/animeganv2_hayao.pdparams to C:\Users\raymo\.cache\ppgan\animeganv2_hayao.pdparams

This step is very slow. I think we should consider somehow mirroring this or host this.... @ryanloney ?

raymondlo84:
This step takes a pretty long time for larger images. I think we should do a print statement to warn the user here when this step is running. I thought it was crashed....
```
# Step 3. Do inference. Setting the model to `.eval()` and using `.no_grad()`
predictor.generator.eval()
with paddle.no_grad():
    result = predictor.generator(input_tensor)
```

raymondlo84:
IMG_3744
IMG_3744_anime_ir
IMG_3744_anime_pg

Interestingly I got different results with my own image. IR one seems nicer :) Completely objective but I think we are doing something to the model and so the results may vary! So would worth a little warning at the end.

raymondlo84:
Last thing I notice is the 'huge' memory spike with the PaddlePaddle model (ti used up 16GB of my ram with a slightly larger image). With the IR we are running much flatter line on the memory usage. This is something worth looking into!

ryanloney:
I'm getting stuck here Python 3.7.9 on Windows 10. 
image

helena-intel:
> This step takes a pretty long time for larger images or even crashed it. I think we should do a print statement to show the user here when this step is running. Maybe we can do a resize just in case?
> 
> ```
> # Step 3. Do inference. Setting the model to `.eval()` and using `.no_grad()`
> predictor.generator.eval()
> with paddle.no_grad():
>     result = predictor.generator(input_tensor)
> ```

Yes, this is what I mentioned yesterday as the known issue. I'll change the resizing.

helena-intel:
> Last thing I notice is the 'huge' memory spike with the PaddlePaddle model (ti used up 16GB of my ram with a slightly larger image). With the IR we are running much flatter line on the memory usage. This is something worth looking into!

This is the same known issue with larger images, will be fixed.

helena-intel:
> I'm getting stuck here Python 3.7.9 on Windows 10.
> image

That is strange, it doesn't break in the CI there. Does it work if you restart the kernel?

helena-intel:
> Interestingly I got different results with my own image. IR one seems nicer :) Completely objective but I think we are doing something to the model and so the results may vary! So would worth a little warning at the end.

It's possible that this is because of the resizing too, and it will be visually the same when the sizes are equal.

ryanloney:
> > I'm getting stuck here Python 3.7.9 on Windows 10.

> 
> That is strange, it doesn't break in the CI there. Does it work if you restart the kernel?

Yes, it still happens after restarting the kernel. I see this: 
```
Error: Can not import avx core while this file exists: C:\Users\ryanlone\code\repos\openvino_env\lib\site-packages\paddle\fluid\core_avx.pyd
```

helena-intel:
> > > I'm getting stuck here Python 3.7.9 on Windows 10.
> 
> > That is strange, it doesn't break in the CI there. Does it work if you restart the kernel?
> 
> Yes, it still happens after restarting the kernel. I see this:
> 
> ```
> Error: Can not import avx core while this file exists: C:\Users\ryanlone\code\repos\openvino_env\lib\site-packages\paddle\fluid\core_avx.pyd
> ```

Do you have Visual C++ redistributable installed? I hope that that is not required :-(

raymondlo84:
> > Interestingly I got different results with my own image. IR one seems nicer :) Completely objective but I think we are doing something to the model and so the results may vary! So would worth a little warning at the end.
> 
> It's possible that this is because of the resizing too, and it will be visually the same when the sizes are equal.

Yes, I think that's the root cause after I review the code further.

helena-intel:
> > Error: Can not import avx core while this file exists: C:\Users\ryanlone\code\repos\openvino_env\lib\site-packages\paddle\fluid\core_avx.pyd
> > ```
> 
> Do you have Visual C++ redistributable installed? I hope that that is not required :-(

I uninstalled C++ and got the same error on Windows. I tried some other paddlepaddle install options, but no luck so far. Apparently GitHub's CI Windows also includes C++, so we need a solution for that. 

What do you think about adding a note that C++ is required? (screenshot for Windows, could add macOS and generic Linux message).
image

notebooks/206-vision-paddlegan-anime/206-vision-paddlegan-anime.ipynb
```diff
@@ -0,0 +1,614 @@
...
+   "id": "03050508-a108-47fd-a95b-7a9f9e6c84a0",
+   "metadata": {},
+   "source": [
+    "There are three transforms: resize, transpose, and normalize, where normalize uses a mean and scale of [127.5,127.5,127.5]. \n",
+    "\n",
+    "The ResizeToScale function is called with (256,256) as argument for size. Further inspection of the function shows that this is\n",
+    "the minimum size to resize to. The ResizeToScale transform resizes images to the size specified in the ResizeToScale params, with width and height\n",
+    "as multiples of 32.\n",
+    "\n",
+    "Now that we know the mean and standard deviation values, and the shape of the model inputs, we can call Model Optimizer and convert the model to IR with these values. "
```
KamilaBanecka, line 326: I got lost here. I understand that the function run() calls the transforms subsequently. Is that right? What args do they need?
for resize I understand that the width and height, where the minimum value is (256,256) 
for transpose I don't see any description
for normalize we need mean and scale [127.5, 127.5, 127.5] why the scale has 3 args?

notebooks/206-vision-paddlegan-anime/206-vision-paddlegan-anime.ipynb
```diff
@@ -0,0 +1,614 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ae08c137-aeda-432a-839c-91c56264e8e7",
+   "metadata": {},
+   "source": [
+    "**Convert Model to IR with Model Optimizer**"
```
KamilaBanecka, line 334: **Skip this comment, if redundant:** 

Can we specify data_type here?

notebooks/206-vision-paddlegan-anime/206-vision-paddlegan-anime.ipynb
```diff
@@ -0,0 +1,614 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "adc56b11",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Step 1. Load the image and convert to RGB\n",
+    "image = cv2.cvtColor(cv2.imread(str(image_path), flags=cv2.IMREAD_COLOR), cv2.COLOR_BGR2RGB)\n",
+    "\n",
+    "# Step 2. Transform the image (only resize and transpose are still required)\n",
```
KamilaBanecka, line 473: **Skip if redundant**

I understand that the images are transformed during model conversion to the IR format, and before running the inference?

ryanloney:
Yes, please add the message about C++ requirement for PaddlePaddle. Installing the package on my Windows laptop and it works. Also tested on my MacBook with Intel CPU and it works.

ryanloney (approved):

notebooks/206-vision-paddlegan-anime/206-vision-paddlegan-anime.ipynb
```diff
@@ -0,0 +1,614 @@
...
+   "id": "f5172a4a-f6ef-4142-8917-6c877706b8d6",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "start = time.perf_counter()\n",
+    "exec_net.infer(inputs={input_key: input_image})\n",
+    "end = time.perf_counter()\n",
+    "time_ir = end - start\n",
+    "print(\n",
+    "    f\"IR model in Inference Engine/CPU: {time_ir:.3f} \" f\"seconds per image, FPS: {1/time_ir:.2f}\"\n",
```
ryanloney, line 543: Can we have a GPU performance check too? Maybe commented out like the PyTorch fastseg. I switched to using GPU device and it works. This is the performance (but still says "CPU" in the printed text)
image

notebooks/206-vision-paddlegan-anime/206-vision-paddlegan-anime.ipynb
```diff
@@ -0,0 +1,614 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "7607ce35-db52-4e1c-add7-8abed748de6a",
+   "metadata": {},
+   "source": [
+    "# Photo to Anime with PaddleGAN and OpenVINO\n",
+    "\n",
+    "This notebook demonstrates converting a [PaddlePaddle/PaddleGAN](https://github.com/PaddlePaddle/PaddleGAN) AnimeGAN model to OpenVINO IR format, and shows inference results on the PaddleGAN and IR models.\n",
+    "\n",
+    "https://github.com/PaddlePaddle/PaddleGAN/blob/develop/docs/en_US/tutorials/animegan.md"
```
ryanloney, line 12:
```suggestion
    "https://github.com/PaddlePaddle/PaddleGAN/blob/develop/docs/en_US/tutorials/animegan.md\n",
    "\n",
    "<img src=\"https://user-images.githubusercontent.com/15709723/123559130-04550100-d74f-11eb-819c-a02284654428.jpg\">"
```

ryanloney: I suggest adding a before and after image to the top, to visually show what the model does

ryanloney (approved):

PR #135, opened 2021-06-29
2021.4

helena-intel:
- use %sx instead of ! for some notebooks
-  remove deprecated legacy resolver from pip install
- remove version pinnings for numpy and importlib-metadata
- pin tensorflow to 2.5


helena-intel:
I pinned TensorFlow to 2.5 since a) I tested that that works and b) that is required for the ONEDNN training speedup in the 301 notebook. Lower versions will probably work just fine, we can look into relaxing this pinning after testing more.

PR #136, opened 2021-06-30
Remove wheel files

ryanloney:
I don't think these wheel files were meant to be uploaded. Probably left over from validation. Removing them.


ryanloney:
@helena-intel feel free to undo this if there is some reason why the wheel files should remain in the main branch.

ryanloney:
CI is passing, I'm going to merge this since it is not changing any code.

helena-intel:
Thanks Ryan! These files should not have been in the main branch indeed.

PR #138, opened 2021-06-30
Fix and future proof broken links to Open Model Zoo model documentation

helena-intel:
Change model documentation links from OMZ Github to docs.openvinotoolkit.org


ryanloney (approved):
I checked all of the URLs and they work for me. Thank you for using the docs site.

PR #139, opened 2021-06-30
Update Docker test and fix Windows line endings in Dockerfile

helena-intel:
- clone notebooks in test script
- also revert Pipfile to Python 3.7. I will change this to 3.8 later.


ryanloney (approved):
The build was successful on my Windows laptop and MacBook. The macOS build took like 30 minutes, but maybe some issue on my end. Windows took 13 minutes. I tried building and testing the image on OpenShift Data Science, and it works there as well. I'm still running into the issue with cloning the repo when it has already been modified in the user space. We should tackle that separately, and I will share more details with you about the issue.

ryanloney:
Actually, I was wrong. Just needed to clear the volume on OpenShift Data Science cluster. It works now with the PRELOAD check! Thank you @helena-intel

PR #140, opened 2021-07-01
Add Model Tools notebook

helena-intel:
Model Downloader, Converter, Info Dumper and Benchmark App.

### DIFF
--- a/notebooks/104-model-tools/104-model-tools.ipynb
+++ b/notebooks/104-model-tools/104-model-tools.ipynb
@@ -1,0 +1,446 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "artificial-discretion",
+   "metadata": {},
+   "source": [
+    "# Working with Open Model Zoo Models\n",
+    "\n",
+    "## Model Downloader, Model Converter, Info Dumper and Benchmark Tool"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "usual-karen",
+   "metadata": {
+    "tags": [
+     "hide"
+    ]
+   },
+   "source": [
+    "This demo shows how to download a model from Open Model Zoo, convert it to OpenVINO's IR format, show information about the model, and benchmark the model.\n",
+    "\n",
+    "| Tool             | Command           | Description                                                                       |\n",
+    "|------------------|-------------------|-----------------------------------------------------------------------------------|\n",
+    "| Model Downloader | `omz_downloader`  | Download models from Open Model Zoo                                               |\n",
+    "| Model Converter  | `omz_converter`   | Convert Open Model Zoo models that are not in OpenVINO's IR format to that format |\n",
+    "| Info Dumper      | `omz_info_dumper` | Print information about Open Model Zoo models                                     |\n",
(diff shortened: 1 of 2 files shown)


raymondlo84:
Should put outputs if GPU not found? Wonder if it's worth the time to list out the 'available devices' so these commands then makes sense.

helena-intel:
> Should put a warning output if GPU not found? Wonder if it's worth the time to list out the 'available devices' so these commands then make sense.

Yes, I wasn't sure how to go about this yet. In the 301 notebook I do show a message. I can add the "else print" part here too, but I also like that this allows you to see more benchmarks on one screen. It's actually a simple command and the cells are already pretty long.

I could make a benchmark function that wraps benchmark app so you can call it with one line and it shows only the relevant output?

helena-intel:
I would like input on default model and cache directories. They are not required for model downloader and converter, but I think it is useful to add them. If we make new OMZ notebooks, we can specify the same directories, and users will not end up with lots of different OMZ model directories on their system, and if they already downloaded the model it will not need to be downloaded again. I made two directories in $HOME in the notebook, but open to other suggestions.

ryanloney:
I am happy with using $HOME for now. Let's see what feedback we get from field. I don't see any problem with using $HOME and even in a container or on cloud this should be fine. 

@helena-intel is it crazy to ask for omz_model_quantizer in this notebook? Or should we put it in a separate one?

ryanloney (approved):

helena-intel:
> I am happy with using $HOME for now. Let's see what feedback we get from field. I don't see any problem with using $HOME and even in a container or on cloud this should be fine.

Yes, to maybe clarify, it doesn't require $HOME to be set, it uses Python's .expanduser(). I have never noticed technical issues. But some people might not love it if we "clutter" their home directory - but then they can change the path in the notebook. We could also use something like $APPDATA on Windows and a hidden directory on Linux, but these directories can become large, I prefer to have them visible.

> @helena-intel is it crazy to ask for omz_model_quantizer in this notebook? Or should we put it in a separate one?

That makes it more complicated because it requires downloading of very large datasets, with licenses that would not allow everyone to use them. Quantization can also take quite a long time. Even if we make it an optional step, it is still not ideal to have it in this simple notebook in my opinion.

ryanloney:
That makes sense. We can have a separate notebook for omz quantizer that downloads a dataset or something, in the future.

helena-intel:
I changed the benchmark_app method: made a function that wraps benchmark app and a GPUNotFoundAlert class that shows a warning if there is no GPU but the benchmark method requires GPU. I can make that more generic later.

image
image

helena-intel:
I added the Alert class to notebook_utils, but also added a fallback class to the import for someone who does not have notebook_utils. It adds some noise to the import section and  we'll have standalone notebooks soon too, so it may be better to remove that. 

```
# Import an error checking function and define a backup class if notebook_utils is not available
sys.path.append("../utils")
class SimpleGPUNotFoundAlert(object):
    def __init__(self):
        print("Running this cell requires an integrated GPU, which is not available on this system.")
        
try:
    from notebook_utils import GPUNotFoundAlert
except:
    GPUNotFoundAlert = SimpleGPUNotFoundAlert
```

ryanloney:
@helena-intel  I tried it again with the changes. Looks good. Please merge.

Issue #144, opened 2021-07-02
Flower Training Notebooks Error


raymondlo84:
I got this error when I run the training notebooks and somehow this line is returning error. I'm on Windows 11, Python 3.7, and just did the installation fresh. 

img_array = keras.preprocessing.image.img_to_array(img)

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-32-e39202bdee35> in <module>
      6 )
      7 print(img)
----> 8 img_array = keras.preprocessing.image.img_to_array(img)
      9 img_array = tf.expand_dims(img_array, 0) # Create a batch
     10 

~\Documents\openvino_env\lib\site-packages\tensorflow\python\keras\preprocessing\image.py in img_to_array(img, data_format, dtype)
    226       dtype = backend.floatx()
    227     kwargs['dtype'] = dtype
--> 228   return image.img_to_array(img, data_format=data_format, **kwargs)
    229 
    230 

~\Documents\openvino_env\lib\site-packages\keras_preprocessing\image\utils.py in img_to_array(img, data_format, dtype)
    307     # or (channel, height, width)
    308     # but original PIL image has format (width, height, channel)
--> 309     x = np.asarray(img, dtype=dtype)
    310     if len(x.shape) == 3:
    311         if data_format == 'channels_first':

~\Documents\openvino_env\lib\site-packages\numpy\core\_asarray.py in asarray(a, dtype, order)
     81 
     82     """
---> 83     return array(a, dtype, copy=False, order=order)
     84 
     85 

TypeError: __array__() takes 1 positional argument but 2 were given
```

raymondlo84:
I then skipped that step and finish the model optimizer and got my xml and bin.

raymondlo84:
I think this may have to do with the PaddleGAN as it had overwritten the numpy and other packages. It actually tried to install older version of numpy and then it broke the Tensorflow 2.5.0. I install numpy 1.19.5 and gast again and it is back to normal! 

In short, the PaddleGAN may need more testing and I think we may need to put that in a separate virtual environment as it conflicts with lots of things... :|

-ray

raymondlo84:
I think it maybe the Keras nightly got a new bug? :\

PR #145, opened 2021-07-02
Fix tensorflow notebook issue

helena-intel:
Pin Pillow to 8.2 to fix issue with Tensorflow Flowers notebook.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,3 +27,4 @@ jupyterlab
 ipython==7.10.*
 jedi==0.17.2
 setuptools>=56.0.0
+Pillow==8.2.*


helena-intel:
@raymondlo84 can you check if this fixes #144 ?

raymondlo84:
Wow... How can you find that so quickly... gosh. Yes, it works now! ... and how how how how how.... 👍

raymondlo84 (approved):
Validated the fix. Will go through CI next.

helena-intel:
> Wow... How can you find that so quickly... gosh. Yes, it works now! ... and how how how how how.... 👍

I wish I could take credit but I figured if a dependency suddenly breaks other people will have run into it and hopefully found a fix and they did. see https://github.com/python-pillow/Pillow/issues/5571
I pinned Pillow to 8.2.* now because I wanted to fix this with a known good version. I'll change it to <8.3 or !=8.3.0 later.

PR #146, opened 2021-07-03
Change TensorFlow and ONNX requirements installation

### LINKED ISSUE #142: Error @ PaddleGAN pip install after Tensorflow 2.5.0 added
The "pip install" now returned this error/warning. Is that normal? 

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
tensorflow 2.5.0 requires gast==0.4.0, but you have gast 0.3.3 which is incompatible.
```

helena-intel:
Install TensorFlow requirements with openvino-dev[tensorflow,onnx] and remove the TensorFlow and ONNX model optimizer requirements from requirements.txt. This installs 2.4.*. The benefit is that these are the supported versions for openvino-dev 2021.4 and less likely to cause issues. It also fixes  #142 

Downside is that for now the ONEDNN speedup in the training notebook does not work anymore.


ryanloney (approved):

ryanloney:
Can we tag a 2021.4 release after this gets merged? Or any other changes to lock in?

requirements.txt
```diff
@@ -0,0 +1,23 @@
+openvino-dev[onnx,tensorflow2]==2021.4
```
ryanloney, line 1:
```suggestion
openvino-dev[onnx,tensorflow2]~=2021.4
```
Since 2021.4 is an LTS release, and since we plan to tag a notebook release, we should use `~=` so security patches/bug fixes in 2021.4.X will be available to someone who uses this requirements file in the future.

ryanloney (approved):

requirements.txt
helena-intel: Good point. `~=` pins major versions though, so `~=2021.4` is also satisfied by 2021.3. I'll change it to `==2021.4.*`

helena-intel:
> Can we tag a 2021.4 release after this gets merged? Or any other changes to lock in?

I'll make a 2021.4 release this week.

PR #147, opened 2021-07-03
Improve Pipfile for Docker image

helena-intel:
Installs openvino-dev with tensorflow and onnx extras, as in https://github.com/openvinotoolkit/openvino_notebooks/pull/146
This installs TF 2.4.2.

It is no longer necessary to install tensorflow-serving-api separately in the Dockerfile.

I installed the PaddleGAN requirements in the image, because otherwise the image is not deterministic anymore (installing in the notebook may override other dependencies).  I also already installed the requirements for Facies Segmentation. I removed some version pinnings from Pipfile (to let pipenv figure them out) and sorted the requirements there.

Renamed .openshift to .docker because image is not limited to OpenShift


helena-intel:
@ryanloney  I noticed that my branch was behind main. I rebased to main now. The large amounts of files changed is just the rename from .openshift to .docker. The only files with content changes are Pipfile, Pipfile.lock and Dockerfile.

ryanloney (approved):
Looks good. I tried building on OpenShift cluster and no issues. Too bad about TF 2.5 and oneDNN, but maybe we can have a pip upgrade optional step in the training notebook? Not critical right now though.

helena-intel:
> Too bad about TF 2.5 and oneDNN, but maybe we can have a pip upgrade optional step in the training notebook? Not critical right now though.

We can discuss, but I prefer to have one supported configuration that we can test and that we know has no version conflicts. In my opinion faster training in one notebook is not worth moving away from that.
I hope this will solve itself in the near future.

PR #159, opened 2021-07-08
Add Chinese language README

ryanloney:
Moved this from the wiki to a MD in the repo, similar to PaddlePaddle git repos. There is now a link at the top of README.md that will load README_ch.md and same in the README_ch.md which will go back to English readme. 

Please let me know if you would like to change this in any way.


raymondlo84:
If we have to be 100% correct then we may want name it README_cn.txt

CN = China,  and zh = simplified chinese =) 

zh_CN 

https://docs.oracle.com/cd/E19683-01/806-6642/new-tbl-72/index.html

helena-intel:
Thanks Ryan! I made Raymond's suggestion (thanks!) and will merge now.

PR #160, opened 2021-07-09
Add Binder option

helena-intel:
Add .binder folder with files that make the notebooks work on mybinder.org See example: https://mybinder.org/v2/gh/helena-intel/openvino_notebooks/86e5366836d09794b3ed1628d46bb3b2778b090f 
I deleted PyTorch/ONNX, PaddleGAN and TensorFlow training from the notebooks in Binder, because MyBinder offers 2GB of memory and that's not enough for these notebooks (I am kind of amazed that the other notebooks work well). I can modify PaddleGAN (in a future PR) to work here - I think that is worth it. Other two notebooks need more memory, sadly. 
mybinder.org also supports JupyterLab, but that requires more memory than Jupyter Notebook and is therefore not recommended (the U2Net notebook doesn't work in Jupyter Lab for example).

The image is automatically built the first time someone tries to open the binder page after a repository update. That takes quite a long time, but after the first times it should be fast (the image is cached globally so it is only slow for the first users after each repo update). I can look into automating the image caching in the future.

We can have launch buttons per notebook in the README like: 
[Binder](https://mybinder.org/v2/gh/helena-intel/openvino_notebooks/feature-binder?filepath=notebooks%2F205-vision-background-removal%2F205-vision-background-removal.ipynb)


ryanloney:
This URL did not work https://mybinder.org/v2/gh/helena-intel/openvino_notebooks/86e5366836d09794b3ed1628d46bb3b2778b090f
image

But the "Launch Binder" button works great. And I was able to run the background removal notebook using about 1GB of RAM. I like how they include the X/2GB of RAM at the top right. Very cool. Also amazing that *most* of the notebooks can run with <2GB of RAM on a free service. Love it.

ryanloney (approved):

helena-intel:
Apologies for the first broken link. I will merge this and make links and launch binder buttons per notebook. I also like the memory usage display. There are a few extensions that do that too. See https://github.com/jtpio/jupyterlab-system-monitor It requires some configuration, but is useful for development.

PR #168, opened 2021-07-12
Add table to notebooks readme 

ryanloney:
Added a list of notebooks in the form of a table to the notebooks/README.md file. Please share any suggestions. I would like to improve the preview images, but not sure where to host them yet. If you have better images to use, please make suggestions.


helena-intel:
Thank you Ryan! This is really great to have. I am going to merge this now. It would be nice to link to this from the main README, preferably somewhere at the top IMO. It should still be clear that users need to follow the installation guide though - that the notebooks do not work standalone. But maybe we can add a bit here
image
that says that we have an API reference, guides for converting/optimizing PyTorch/TensorFlow/PaddlePaddle models, and model demos - click here for a full list of included notebooks?

helena-intel:
Thanks Ryan!

helena-intel (approved):

PR #170, opened 2021-07-13
Make AnimeGAN inference optional on PaddlePaddle model

helena-intel:
Add PADDLEGAN_INFERENCE setting. Default True, if set to False, inference on PaddlePaddle model will be skipped. This allows running the notebook on systems with limited resources, and speeds up "Run All Cells" if you are not interested in comparisons with the original PaddleGAN model but just want to quickly show the notebook.

This allows the notebook to run on Binder too.

Also added `--log_level = CRITICAL` to MO to prevent showing some errors that are not relevant for this demo.


ryanloney (approved):
Just tested this with and without the ENV setting. Everything works for me

PR #173, opened 2021-07-14
Add note about C++ to installation guide

helena-intel:
openvino-dev recently started requiring C++ because it is required by a dependency that is used for a few supported models. 
Add a note about this to the main README and an extra step in the Windows Guide https://github.com/openvinotoolkit/openvino_notebooks/wiki/Windows


helena-intel:
We already had a note in the troubleshooting section of our Windows guide - but it seems better to be a bit more up-front about this being a requirement.

ryanloney (approved):

PR #179, opened 2021-07-21
104: support OMZ models with multiple model files

helena-intel:
Solves issue that benchmarking part did not work for models with multiple model files.  Now when there are multiple models, the first is chosen, and a warning message is displayed.
image

Also formatted the code, and removed the backup function for the warning/alert class. I added that in the first version for if notebook_utils was not available, but it adds clutter and seems unnecessary.


ryanloney (approved):
Thanks for updating this

PR #186, opened 2021-07-27
Add badges

alexeyhorkin:
I thought it would be nice to add badges like https://github.com/openvinotoolkit/openvino in REAMDE.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,6 @@ English | [简体中文](README_cn.md)
 
 # 📚 OpenVINO™ Notebooks
 
-![GitHub release (latest by date)](https://img.shields.io/github/v/release/openvinotoolkit/openvino_notebooks)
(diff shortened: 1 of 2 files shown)


alexeyhorkin:
Looks like this:
image

helena-intel:
Thanks @alexeyhorkin I like this a lot! I'm not sure yet about the release badge - we have been talking about how to go about releases. The 0.1.0 release is old. Currently we encourage users to use the main branch - that is the version that is tested and supported. We also do not yet expose how to use the Docker image to users - but it would be good to offer that as an option and then having this badge is great.

helena-intel (approved):
Thanks @alexeyhorkin ! If we do decide to make proper releases we can always put the release badge back. I really like the badges.

ryanloney (approved):
Thanks @alexeyhorkin! These badges look nice. 

Helena - I think we should tag a 2021.4 release soon.

PR #187, opened 2021-07-27
Add Notebook Utils notebook

helena-intel:
Add a notebook with utility functions, for downloading files (with a progress bar), visualizing segmentation output, etc. Some of these functions currently exist in notebook_utils.py. Having them in a notebook is nice - it allows you to quickly try them out, see the effects of different parameters, etc. It also allows for having tests in the notebook. 

If a new PR is created that modifies notebook_utils.ipynb, a CI job will automatically generate the corresponding notebook_utils.py file. It will then make a PR to the branch that the original PR is from, so the person who made the PR to main can have a look at it and accept it in their branch, which will then be updated with the generated notebook_utils.py file. This way, the files are always in sync. 

There is not yet a CI test that works the other way around: if someone updates the notebook_utils.py file, ideally the CI would fail with a message to update the notebook instead. I tried a few things but that was not trivial so I want to merge this first. The CI runs whenever notebook_utils.ipynb or notebook_utils.py is changed, so there is no risk that notebooks break because of changes in either of these files.

I plan to expand this notebook - we may have a series at some point. Already planned: 1) resize images with or without adding padding (and resizing padded images back to original size), 2) A generic class that abstracts away most of the boilerplate for video inference (like in the monodepth notebook), 3) Functions/classes to deal with async and live inference.

We should add a section about this to the CONTRIBUTING guide too - but that can wait after the other PR that modifies that is merged.


helena-intel:
I notice a CI error - this passed in my branch, but it looks like it assumes the source branch to exist in this repo, which it doesn't. I'll fix that.

helena-intel:
I removed the CI function for now. I will have a look later to make this work with PRs from a fork. For now we can convert manually. If someone updates the py or ipynb file and does not update the other file, the files are out of sync. This should not happen, but is also not a major issue. The CI will catch errors, and we are aware that these files should be in sync. 

The command to convert the notebook to the script is 
`jupyter nbconvert notebook_utils.ipynb --TagRemovePreprocessor.remove_cell_tags=hide --to script`

The notebook includes the tag "hide" for cells that should not be included in the script (the test cells). You can see/set them in Jupyter Lab, usually at the right of the window:
image

I also added a `check_openvino_version()` function in the latest commit. This can be added to the imports cell to check with one line of code if the expected OpenVINO version is installed. We could also not do that and just assume 2021.4 for now - but I found the lines of code that check for this at the start of e.g. 103 a bit distracting. In the future we could also integrate a `check_install` function at the top that checks more things - but that requires more testing to work reliably and cross platforms/python versions.

ryanloney (approved):

PR #188, opened 2021-07-28
Rename U2Net model file

helena-intel:


helena-intel:
Fixing issue with downloading weights file from Google drive

PR #201, opened 2021-08-03
207 paddlegan superresolution

helena-intel:
Superresolution notebook with PaddleGAN model. Works very well with some images, not as amazing with others, but it doesn't make them worse (unlike some models). I tried it with old scanned pictures, and lower-res versions of higher res pictures (e.g. shared on Facebook). In the notebook I used a 400x300 Coco image from Flickr. It works very well for that.

On my TGL laptop, PaddlePaddle inference takes 115 seconds, OpenVINO inference 14 seconds.
On iGPU inference takes 2.5 seconds, but it takes 3 minutes to load the network (great opportunity for model caching!)

README and notebook intro could be a bit expanded. I have not added Binder links yet, but it does work: https://mybinder.org/v2/gh/helena-intel/openvino_notebooks/207-paddlegan-superres?filepath=notebooks%2F207-vision-paddlegan-superresolution%2F207-vision-paddlegan-superresolution.ipynb

I doubted whether to make a new folder or put it in the existing superresolution folder. I think this is clearer because it is a very different model and notebook - but if you want to move it to the other folder that is okay too.

@raymondlo84 @ryanloney please make any changes you want.


raymondlo84:
We forgot to save the final result actually. Here is the patch to save the result as png.

```
result_pil = Image.fromarray(image_super)
bicubic_pil = Image.fromarray(image_bicubic)
gif_image_path = OUTPUT_DIR / Path(IMAGE_PATH.stem + "_comparison.gif")
final_image_path = OUTPUT_DIR / Path(IMAGE_PATH.stem + "_super.png")


result_pil.save(
    fp=str(gif_image_path),
    format="GIF",
    append_images=[bicubic_pil],
    save_all=True,
    duration=1000,
    loop=0,
)
result_pil.save(fp=str(final_image_path), format="png")
DisplayImage(open(gif_image_path, "rb").read(), width=1920 // 2)
```

ryanloney (approved):

PR #212, opened 2021-08-23
Add rst generating to workflow and fix headers

Debskij:


helena-intel (changes requested):
Thanks for this. I have one small change request. Since it has been 10 days, could you also rebase/merge your PR to main? 

Note though that currently the `convert_notebooks.sh` script makes some changes to the notebooks to execute much faster. Only a few epochs of the 301 notebook are run, and for superresolution only a few frames are processed.  So far these converted files are used to facilitate converting the notebooks to other languages and to have a quick overview of what the executed notebooks look like. If these rst files will be used for documentation, we should run the notebooks unaltered - but that may cause issues with Github's public CI run time limit. I suggest leaving as is for now, but this should be addressed before the documentation goes live.

PR #219, opened 2021-08-31
Pin black version to prevent PyPI resolver issues

helena-intel:
New version of black, 21.8b0, requires typing-extensions>3.10, which is not compatible with Tensorflow, which requires ~3.7.4. This results in the PyPI resolver trying forever to find compatible requirements. This PR solves that by pinning black to the previous version, 21.7b0


ryanloney (approved):
Approving so the CI checks kick off. Thanks for fixing this

PR #220, opened 2021-08-31
Run CI on push to main, add 103 back to CI

helena-intel:
We currently only run the CI on PR. This is fine when there are few PRs and the process is quick, but things can change between PR approval and merge. Adding a check to run the CI on merge to main.
Also added 103 notebook back to CI. This was removed in #214. Discussed with @Debskij


ryanloney (approved):

ryanloney (approved):
This is a much better solution to the random download failures. Thank you!

PR #227, opened 2021-09-02
Add weights to read_model in 004

helena-intel:
004 adds quite a bit of boilerplate to check if model files already exist. If read_network() is called with just a model parameter and the weights file does not exist, the Jupyter kernel crashes (this only happens if the user deleted the model bin file). But if read_network() is called with both model and weights parameters, an exception is raised if either file doesn't exist, and we can skip all the checking code. Especially for a "hello" notebook, that makes it friendlier to read, and it is more consistent. 

We should revisit the other notebooks too, but to prevent annoying merge conflicts I will not do that all in this PR. 

Code now looks like this:
image

### DIFF
--- a/notebooks/004-hello-detection/004-hello-detection.ipynb
+++ b/notebooks/004-hello-detection/004-hello-detection.ipynb
@@ -2,76 +2,78 @@
  "cells": [
   {
    "cell_type": "markdown",
+   "id": "01412caf",
+   "metadata": {},
    "source": [
     "# Hello World Text Detection\n",
     "\n",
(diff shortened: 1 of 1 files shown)


ryanloney:
I tried removing the .bin and the error is very clear:
Screen Shot 2021-09-04 at 13 03 34

Same if I just remove the .xml (or both files): 
Screen Shot 2021-09-04 at 13 04 35

Looks good to me as long as the CI passes. Thank you for updating the notebook.

ryanloney (approved):

PR #228, opened 2021-09-02
Webcam Hello World

raymondlo84:
Here is a webcam version of hello world. It uses the same model as the 001-hello-world but we use a webcam feed as the input. The main issue is how we can do CI with this at all, and that's why I'm thinking we have to put this under a 4xx series as it will have hardware dependencies.

However, I will push a pull request here and see what we think about it and so at least I have this somewhere. :)


raymondlo84:
scrnli_9_2_2021_1-16-17 PM
Not sure how can pass the CI now, but here is the sample output ...

raymondlo84:
Good points. Should I switch the model to SSD? This one will give a bounding box and so we are not getting random results as easily. The image classification one is classifying the entire screen, so often you have to get very close to the object... :D Maybe we shall make a little crop box and so people can put things inside the box to classify instead of the entire video... 

https://docs.openvinotoolkit.org/latest/omz_models_model_mobilenet_ssd.html

Let me make all updates next week, and let's not merge it. I want to kick off this thread so it will get done eventually. ;)

-ray

> Thanks @raymondlo84 this is cool to have! I am not yet very enthusiastic about the model though. I see your demo image is in a dark room, and probably with a good webcam. When I tried it the results were kind of meh, even when I pointed the camera so that there wasn't much in the background
> image. When I try it with just my face it keeps recognizing a shower cap, seat belt, or oxygen mask. None of which I am wearing :-)
> Have you tried other models too?
> 
> My laptop's fan did start making noise, but the performance seemed pretty fast. Should we display the fps?
> 
> For the CI, we can patch the notebook so that the CI runs it on a video. It is nice that the last cell already says how to run the demo with a video. If we provide a sample video, we would only need to uncomment that, and comment the webcam line. That would not test if the webcam part actually works, but we will notice quickly if OpenCV completely changes and that stops working. We should manually test that the webcam works on all supported OS's, but we have done that before with the hand pose demo so I do not expect issues.
> 
> Some comments:
> 
> * If you do `pip install .ci\dev-requirements.txt` you get jupyterlab-code-formatter. You can then click on the "format code" button and it will fix some minor formatting issues, like adding a space after a `#` etc.
> * I like the comments in the video processing function, but would like to have a markdown cell with a high level overview of what happens too. We should also explain the video index parameter, that if you have two webcams (for example a front and back camera) you may need to set it to 1.
> * Could we have a solid background for the text? Depending on the environment the text may be hard to read.
> * If I run the inference on a video (uncommenting the last cell and using a random video) the video disappears as soon as it is finished. That's a bit unexpected to me. If I move the `clear_output` line before  `display(i)` the last frame is still shown. It seems to not interfere with the webcam inference. What do you think? Video was also displayed faster than it's frame rate (because inference is super fast!), but that seems fine for this notebook.
> 
> In the future I would like to use https://ipywebrtc.readthedocs.io/en/latest/CameraStream.html . That also works on Docker images. And it has some friendly methods to select, for example, the "front facing camera". But that should be easy to swap, or we can still keep this as the opencv webcam notebook.

ryanloney:
@raymondlo84 yes, please can we do the mobilenet-ssd model instead?

raymondlo84:
> @raymondlo84 yes, please can we do the mobilenet-ssd model instead?

I talked to Helena and I will actually make another one with mobilenet-ssd next! :) Let's have this one as a basic 100 hello world webcam as it's very very simple.

raymondlo84:
image
I dropped a shadow by drawing the text twice with a +1 +1 offset :D and now it is super easy to read! ... I think I'm getting smarter and smarter now :D hahaaa

ryanloney (changes requested):
Can you change the camera to default to 0? I've tried on 3 laptops now and everyone has 0 as the built in webcam.

raymondlo84:
> Can you change the camera to default to 0? I've tried on 3 laptops now and everyone has 0 as the built in webcam.

Yes, my bad and I keep forgetting :)!

helena-intel:
@raymondlo84 the webcam video output is flickering very badly on Ubuntu 20.04 for me, same as in #247 What version of Ubuntu did you test? 

There are also a few superfluous inputs/variables and some style issues (that  jupyterlab-code-formatter should be able to fix). See `nbqa flake8 <filename.ipynb>` for  the full output.
image

raymondlo84:
I have noticed the same issue with Ubuntu, and it seems the clear_output is not waiting or doing the double buffer...

https://notebook.community/CestDiego/emacs-ipython-notebook/tests/notebook/nbformat4/Animations%20Using%20clear_output



> @raymondlo84 the webcam video output is flickering very badly on Ubuntu 20.04 for me, same as in #247 What version of Ubuntu did you test?
> 
> There are also a few superfluous inputs/variables and some style issues (that jupyterlab-code-formatter should be able to fix). See `nbqa flake8 <filename.ipynb>` for the full output.
> image

raymondlo84:
I have to put in a workaround as it seems Linux won't sync with the wait=True command. I have put in imshow function but now with a little pop up screen.

adrianboguszewski:
@helena-intel @raymondlo84 
Should this PR be closed as 401 is already merged?

PR #229, opened 2021-09-03
Fix/improve CI issues with 105 and 208

helena-intel:
- Temporarily remove progress bar from download_file in 105: this causes issues on macOS with this large file: Jupyter shows many iopub data rate messages.
- Add --num_attempts to model downloader in 208 to make it a bit more robust and try downloading a model file up to three times if it fails (rest of changes in 208 are metadata changes)
- Set progress bar to display a bit less often in notebook_utils to prevent Jupyter iopub errors (this is not enough to fix it for macOS but a more robust default on Windows and Linux). Also minor update to download_model function to only show progress bar on bin download, not xml (since that is usually very small). notebook_utils.py and notebook_utils.ipynb are now in sync.

This is not a real fix for the download issues with 208. The model source file is stored on Google Drive, and sometimes the download limit is reached.

### DIFF
--- a/notebooks/105-language-quantize-bert/105-language-quantize-bert.ipynb
+++ b/notebooks/105-language-quantize-bert/105-language-quantize-bert.ipynb
@@ -84,7 +84,7 @@
    "cell_type": "code",
    "execution_count": null,
    "source": [
-    "download_file(MODEL_LINK, directory=MODEL_DIR)\n",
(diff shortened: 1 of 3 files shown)


helena-intel:
Created issue to improve progress bar and put it back in 105 when that is done. https://github.com/openvinotoolkit/openvino_notebooks/issues/230

ryanloney (approved):

PR #207, opened 2021-08-07
Draft of 302 pytorch nncf quantization 

ljaljushkin:
The goal of this notebook to demonstrate how to use NNCF 8-bit quantization to optimize the PyTorch model for inference with OpenVINO Toolkit.

PR #239, opened 2021-09-08
Pin Pillow requirement, add nncf to Docker image

helena-intel:
- Pin Pillow in requirements.txt and Pipfile because of security vulnerability reported by snyk #231 
  - We pinned Pillow to 8.2 because of an issue with 8.3.0 - that issue was fixed in 8.3.1.
- nncf will be added to requirements.txt in a #207  - this commit is just to prepare the Docker image. For docker `pipenv lock` needs to be manually run at the moment, so it is useful to have this in the image already.

### DIFF
--- a/.docker/Pipfile
+++ b/.docker/Pipfile
@@ -17,12 +17,13 @@ jupyterhub = "==1.1.0"
 jupyterlab = "*"
 matplotlib = "<3.4"
 nbval = "*"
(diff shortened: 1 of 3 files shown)


helena-intel:
Tested in local CI that Docker image tests passes: 
image

ryanloney (approved):

PR #245, opened 2021-09-09
Update Docker image to 2021.4.1

helena-intel:
Updated Pipfile.lock to OpenVINO 2021.4.1
Also update Docker notebooks test to use patch_notebooks script


ryanloney (approved):
Thanks for updating it

PR #247, opened 2021-09-10
Added pose estimation live demo

adrianboguszewski:
Code review welcome :)


raymondlo84:
notebooks/402-pose-etimation-webcam/model/human-pose-estimation-0001.bin  ..

We have to decide if we shall use the model downloader to get this model. We generally don't keep large binary on the repository if possible.

-ray

raymondlo84:
https://github.com/openvinotoolkit/openvino_notebooks/tree/main/notebooks/104-model-tools 

FYI.

helena-intel:
Thanks for this notebook @adrianboguszewski This looks like a fun demo but when I tried it on Ubuntu 20.04 the output was flickering a lot. I'll try Windows on Monday. Have you or @raymondlo84 tried it on Linux? I tested on a fresh install of Ubuntu, it is possible that I am missing some video optimization libraries or  settings.

I agree we should use Model Downloader for this notebook. It is an Intel model so Model Converter is not necessary, `omz_downloader` is enough. That would also make it easy to allow the user to switch to a different pose model.

ryanloney:
I tried running on a laptop with RHEL 8.4. It works fine except for a lot of flickering. I noticed this the other day when running 401 notebook on OpenSUSE, though it is slightly better on RHEL. I have not tried other Linux versions yet, but seems like this may be a Linux specific issue, as it does not happen for me on Windows and when I've seen Raymond run it on macOS it seems fine too. Maybe there is nothing we can do, but sharing in case you have ideas of tweaks to the config that might help. 

Here is a GIF showing the flickering:
ILZnvl36gi

raymondlo84:
Shouldn't we do the fps this way? 

```
            start_time = time.time()

            # get results
            results = exec_net.infer(inputs={input_key: input_img})
            poses, scores = process_results(frame, results)

            stop_time = time.time()
            # calculate FPS
            fps = 1/(stop_time - start_time);
```

adrianboguszewski:
Thanks for the comments!

That's weird. I use Fedora and it's not flickering for me at all.
I added omz_downloader for downloading the model, but for now, it's not true that the user can switch to a different model (there is completely different postprocessing).

PR #251, opened 2021-09-11
Code style updates

helena-intel:
- black formatting
- minor refactoring to use utils function where useful
- remove check for installation of OpenVINO 2021.4 from some notebooks (is assumed now)
- replaced `/path/to/model_optimizer` with `mo`
- added parameters to many function calls
- modified some descriptions and headers for clarification and consistency
- added semicolons after plt.imshow() calls to prevent printing matplotlib axis info

Not intended to be complete - incremental improvements

More readable diff attached as PDF (created with nbdime and ansi2html)
[diff.pdf](https://github.com/openvinotoolkit/openvino_notebooks/files/7147820/diff.pdf)

### DIFF
--- a/notebooks/002-openvino-api/002-openvino-api.ipynb
+++ b/notebooks/002-openvino-api/002-openvino-api.ipynb
@@ -2,13 +2,16 @@
  "cells": [
   {
    "cell_type": "markdown",
+   "id": "b875b31c",
+   "metadata": {},
    "source": [
     "# OpenVINO API Tutorial"
-   ],
-   "metadata": {}
+   ]
   },
   {
    "cell_type": "markdown",
+   "id": "09abd9f4",
+   "metadata": {},
    "source": [
     "This notebook explains the basics of the OpenVINO Inference Engine API. It covers:\n",
     "\n",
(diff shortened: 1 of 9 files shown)


notebooks/002-openvino-api/002-openvino-api.ipynb
```diff
@@ -2,13 +2,16 @@
  "cells": [
   {
    "cell_type": "markdown",
+   "id": "b875b31c",
```
ryanloney, line 5: @helena-intel  just curious, but why do these ids show up now? Do we use them for something?

ryanloney (approved):
I tried each notebook and reviewed the changes. LGTM

notebooks/002-openvino-api/002-openvino-api.ipynb
helena-intel: This is a change in the notebook format. This is not the best link but gives some info and issues https://github.com/jupyterlab/jupyterlab/issues/9729#issuecomment-803197368

helena-intel:
> I tried each notebook and reviewed the changes. LGTM

Thanks for trying them all!

PR #253, opened 2021-09-15
Added pose estimation live demo

adrianboguszewski:
I fixed a lot of things, added documentation and removed big files from the git history. Hence, I create a new PR.


helena-intel:
Thanks @adrianboguszewski ! This seems to work really well. I still have some Linux issues sadly. I got an error about missing libraries on my newish Ubuntu 20.04 computer:
image
We should document that and maybe wrap the function in a try/except that gives an error message that tells what to install.

After restarting the kernel and running all cells I consistently got 
image
I could see that my webcam light was still on - apparently that had not been properly closed and restarting the kernel from within Jupyter was not enough. After stopping the Jupyter server (ctrl-C in the terminal) and closing my existing Jupyter tabs the notebook worked again.

adrianboguszewski:
The problem is with your compilation of OpenCV. It was compiled without GUI support. It should never happen if you install it via pip.

I think I can better release the resources. Let's see if it improves the user experience.

adrianboguszewski:
I improved documentation and added a nicer image ;) (https://cocodataset.org/#explore?id=146397). Please tell me if I should improve something else.

adrianboguszewski:
@raymondlo84 @ryanloney @helena-intel 
I improved documentation and added a nicer image ;) (https://cocodataset.org/#explore?id=146397). Please tell me if I should improve something else.

ryanloney:
> @raymondlo84 @ryanloney @helena-intel
> I improved documentation and added a nicer image ;) (https://cocodataset.org/#explore?id=146397). Please tell me if I should improve something else.

Unfortunately we cannot use images of people without completing a privacy impact assessment and getting legal approval. Can you find another image without a person?

adrianboguszewski:
> > @raymondlo84 @ryanloney @helena-intel
> > I improved documentation and added a nicer image ;) (https://cocodataset.org/#explore?id=146397). Please tell me if I should improve something else.
> 
> Unfortunately we cannot use images of people without completing a privacy impact assessment and getting legal approval. Can you find another image without a person?

But this is the pose estimation demo. It doesn't work without people. This image comes from the COCO dataset, so I hope everything is licensed properly.

ryanloney:
@adrianboguszewski sorry, I was confusing this with the other webcam demo. Let's sync on Thursday about our options.

helena-intel:
The annotations of COCO are CC licensed, but the COCO consortium does not own the copyright of the images in the dataset, so we would still need to have legal review of this image https://cocodataset.org/#termsofuse

adrianboguszewski:
The image comes from Flickr. Many of them have CC license also (https://www.flickr.com/creativecommons/). Some are in the public domain. So let's maybe find some public domain image with people and use it?

adrianboguszewski:
The frame (https://user-images.githubusercontent.com/4547501/134550328-a5c99d22-ae60-4281-8120-a8f06a17b96a.png) from https://github.com/intel-iot-devkit/sample-videos/blob/master/preview/store-aisle-detection.gif

CONTRIBUTING.md
```diff
@@ -115,6 +115,7 @@ Names should be descriptive but not too long. We use the following numbering sch
 - `100-` OpenVINO tool tutorials: explain how to optimize and quantize notebooks.
 - `200-` OpenVINO model demos: demonstrate inference on a particular model.
 - `300-` Training notebooks: notebooks that include code to train neural networks.
+- `400-` Live demo notebooks: demonstrate inference on live webcam.
```
ryanloney, line 118:
```suggestion
- `400-` Live demo notebooks: demonstrate inference on a live webcam.
```

README.md
```diff
@@ -55,6 +55,14 @@ Tutorials that include code to train neural networks.
 | [301-tensorflow-training-openvino-pot](notebooks/301-tensorflow-training-openvino/) | Use Post-training Optimization Tool (POT) to quantize the flowers model | |
 | [302-pytorch-quantization-aware-training](notebooks/302-pytorch-quantization-aware-training) | Use Neural Network Compression Framework (NNCF) to quantize PyTorch model | |
 
+### Live Demos
+
+Demos working with a webcam.
```
ryanloney, line 60:
```suggestion
Live inference demos that require a webcam.
```

notebooks/402-pose-estimation-webcam/402-pose-estimation.ipynb
```diff
@@ -0,0 +1,488 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "collapsed": false,
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "If you don't have a webcam, you can run with your local video. It could be any file supported in OpenCV (see: https://docs.opencv.org/4.5.1/dd/d43/tutorial_py_video_display.html)"
```
ryanloney, line 450:
```suggestion
    "If you don't have a webcam, you can still run this demo with a video file. Any format supported by OpenCV will work (see: https://docs.opencv.org/4.5.1/dd/d43/tutorial_py_video_display.html)"
```

notebooks/402-pose-estimation-webcam/402-pose-estimation.ipynb
```diff
@@ -0,0 +1,488 @@
...
+    "        print(e)\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Run\n",
+    "\n",
+    "Run with webcam. Your main webcam has number 0. If you have more webcams, they have consecutive numbers. Set flip to True if you're using the front camera. Some graphics drivers (especially Linux open source drivers) cause flickering. If you experience that, set use_popup to True."
```
ryanloney, line 425:
```suggestion
    "Run using a webcam as the video input. By default, the primary webcam is set with `source=0`. If you have multiple webcams, each one will be assigned a consecutive number starting at 0. Set `flip=True` when using a front-facing camera. Some graphics drivers, especially open-source drivers for Linux, may cause flickering. If you experience flickering, set `use_popup=True`."
```

notebooks/402-pose-estimation-webcam/402-pose-estimation.ipynb
```diff
@@ -0,0 +1,488 @@
...
+   "cell_type": "markdown",
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "### Main processing\n",
+    "\n",
+    "Run pose estimation on the specified source. It could be a webcam or video file."
```
ryanloney, line 328:
```suggestion
    "Run pose estimation on the specified source. Either a webcam or a video file."
```

notebooks/402-pose-estimation-webcam/402-pose-estimation.ipynb
```diff
@@ -0,0 +1,488 @@
...
+   "cell_type": "markdown",
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "### Pose drawing\n",
+    "\n",
+    "We draw the poses on the image to visualize them. Joints are drawn as circles and limbs are drawn with lines. Code based on [Human Pose Estimation Demo](https://github.com/openvinotoolkit/open_model_zoo/tree/master/demos/human_pose_estimation_demo/python)"
```
ryanloney, line 277:
```suggestion
    "Draw pose overlays on the image to visualize estimated poses. Joints are drawn as circles and limbs are drawn as lines. The code is based on the [Human Pose Estimation Demo](https://github.com/openvinotoolkit/open_model_zoo/tree/master/demos/human_pose_estimation_demo/python) from Open Model Zoo."
```

notebooks/402-pose-estimation-webcam/402-pose-estimation.ipynb
```diff
@@ -0,0 +1,488 @@
...
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "### Result processing\n",
+    "\n",
+    "A bunch of useful functions to transform results into poses.\n",
+    "\n",
+    "Firstly we pool heatmap. As pooling is not available in numpy, we use a simple trick to do that directly in numpy. Then we use non-maximum suppression to get keypoints from the heatmap. After that, we decode poses using the created decoder. As the input image is bigger than network outputs, we need to multiply all pose coordinates by the scaling factor."
```
ryanloney, line 193:
```suggestion
    "First, we will pool the heatmap. Since pooling is not available in numpy, we use a simple method to do it directly with numpy. Then, we use non-maximum suppression to get the keypoints from the heatmap. After that, we decode poses using the decoder. Since the input image is bigger than the network outputs, we need to multiply all pose coordinates by a scaling factor."
```

notebooks/402-pose-estimation-webcam/402-pose-estimation.ipynb
```diff
@@ -0,0 +1,488 @@
...
+   "cell_type": "markdown",
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "### OpenPoseDecoder\n",
+    "\n",
+    "We need a decoder to transform raw results from the neural network into poses. This magic happens inside Open Pose Decoder. It comes from [OpenVINO Model Zoo](https://github.com/openvinotoolkit/open_model_zoo/blob/master/demos/common/python/models/open_pose.py) and it's suitable for the \"human-pose-estimation-0001\".\n",
```
ryanloney, line 163:
```suggestion
    "We need a decoder to transform the raw results from the neural network into pose estimations. This magic happens inside Open Pose Decoder, which is provided in the [Open Model Zoo](https://github.com/openvinotoolkit/open_model_zoo/blob/master/demos/common/python/models/open_pose.py) and compatible with the \"human-pose-estimation-0001\" model.\n",
```

notebooks/402-pose-estimation-webcam/402-pose-estimation.ipynb
```diff
@@ -0,0 +1,488 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "Input key is the name of the input node and output keys are names of output nodes of the network. In the case of the OpenPose Model, we have one input and two outputs: pafs and keypoints heatmap."
```
ryanloney, line 137:
```suggestion
    "Input key is the name of the input node and output keys contain names of output nodes of the network. In the case of the OpenPose Model, we have one input and two outputs: pafs and keypoints heatmap."
```

notebooks/402-pose-estimation-webcam/402-pose-estimation.ipynb
```diff
@@ -0,0 +1,488 @@
...
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "### Load model\n",
+    "\n",
+    "Downloaded models are located in a fixed structure, which indicates vendor, model name and precision.\n",
+    "\n",
+    "We need only a few lines of code to be able to run the model. Firstly we create an Inference Engine, then we read the network architecture and weights from the files and load it into the desired device."
```
ryanloney, line 104:
```suggestion
    "Only a few lines of code are required to run the model. First, we create an Inference Engine. Then we read the network architecture and model weights from the .bin and .xml files to load onto the desired device."
```

notebooks/402-pose-estimation-webcam/402-pose-estimation.ipynb
```diff
@@ -0,0 +1,488 @@
...
+   "cell_type": "markdown",
+   "metadata": {
+    "collapsed": false
+   },
+   "source": [
+    "### Download model\n",
+    "\n",
+    "We use omz_downloader, which is a utility from openvino-dev package. omz_downloader automatically creates a directory structure and downloads the selected model.\n",
+    "\n",
+    "If you want to download another model, please change the model name and precision. *Note: In that case, you need different pose decoder!*"
```
ryanloney, line 60:
```suggestion
    "If you want to download another model, please change the model name and precision. *Note: This will require a different pose decoder*"
```

notebooks/402-pose-estimation-webcam/402-pose-estimation.ipynb
```diff
@@ -0,0 +1,488 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "collapsed": false
+   },
+   "source": [
+    "### Download model\n",
+    "\n",
+    "We use omz_downloader, which is a utility from openvino-dev package. omz_downloader automatically creates a directory structure and downloads the selected model.\n",
```
ryanloney, line 58:
```suggestion
    "We use `omz_downloader`, which is a command line tool from the `openvino-dev` package. `omz_downloader` automatically creates a directory structure and downloads the selected model.\n",
```

notebooks/402-pose-estimation-webcam/402-pose-estimation.ipynb
```diff
@@ -0,0 +1,488 @@
...
+   "metadata": {
+    "collapsed": true,
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "# Pose estimation with OpenVINO\n",
+    "\n",
+    "This notebook demonstrates live pose estimation in OpenVINO. We use [OpenPose](https://github.com/openvinotoolkit/open_model_zoo/tree/master/models/intel/human-pose-estimation-0001) pose estimation model from [Open Model Zoo](https://github.com/openvinotoolkit/open_model_zoo/). At the bottom you can see live results with your webcam or a video file."
```
ryanloney, line 14:
```suggestion
    "This notebook demonstrates live pose estimation with OpenVINO. We use the OpenPose model [human-pose-estimation-0001](https://github.com/openvinotoolkit/open_model_zoo/tree/master/models/intel/human-pose-estimation-0001) from [Open Model Zoo](https://github.com/openvinotoolkit/open_model_zoo/). At the bottom of this notebook, you will see live inference results from your webcam. You can also upload a video file."
```

notebooks/README.md
```diff
@@ -45,3 +45,10 @@ Tutorials that include code to train neural networks.
...
 | [302-pytorch-quantization-aware-training](302-pytorch-quantization-aware-training) | Use Neural Network Compression Framework (NNCF) to quantize PyTorch model | |
+
+## Live Demos
+
+Demos working with a webcam.
+| Notebook | Description | Preview |
+| :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------: | :------------------------------------------------------------------------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------: |
+| [402-pose-etimation-webcam](402-pose-estimation-webcam/) | Estimate position and orientation of each person | <img src="https://user-images.githubusercontent.com/4547501/134550328-a5c99d22-ae60-4281-8120-a8f06a17b96a.png" width=225> |
```
ryanloney, line 54:
```suggestion
| [402-pose-etimation-webcam](402-pose-estimation-webcam/) | Human pose estimation with a webcam or video file | <img src="https://user-images.githubusercontent.com/4547501/134550328-a5c99d22-ae60-4281-8120-a8f06a17b96a.png" width=225> |
```

notebooks/402-pose-estimation-webcam/README.md
```diff
@@ -0,0 +1,13 @@
+# Pose estimation demo
+
+![pose estimation](https://user-images.githubusercontent.com/4547501/134550328-a5c99d22-ae60-4281-8120-a8f06a17b96a.png)
+
+Pose estimation allows to determine a 2D position and orientation of each person in an image/video. There are body skeletons consisting of 18 predefined keypoints (joints) and 19 connections between them (limbs).
```
ryanloney, line 5:
```suggestion
Pose estimation predicts the 2D position and orientation of each person in an image or video. Skeletons consisting of 18 predefined keypoints (joints) and 19 connections between them (limbs) are visualized as an overlay on the images or video.
```

notebooks/402-pose-estimation-webcam/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Notebook Contents
+
+This notebook demonstrates pose estimation with the [Open Pose model](https://github.com/openvinotoolkit/open_model_zoo/tree/master/models/intel/human-pose-estimation-0001) in OpenVINO.
```
ryanloney, line 9:
```suggestion
This notebook demonstrates human pose estimation with OpenVINO using the OpenPose [human-pose-estimation-0001](https://github.com/openvinotoolkit/open_model_zoo/tree/master/models/intel/human-pose-estimation-0001) model from Open Model Zoo.
```

README.md
```diff
@@ -55,6 +55,14 @@ Tutorials that include code to train neural networks.
...
+| Notebook | Description | Preview |
+| :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------: | :------------------------------------------------------------------------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------: |
+| [402-pose-etimation-webcam](notebooks/402-pose-estimation-webcam/) | Estimate position and orientation of each person | <img src="https://user-images.githubusercontent.com/4547501/134550328-a5c99d22-ae60-4281-8120-a8f06a17b96a.png" width=225> |
```
ryanloney, line 63:
```suggestion
| [402-pose-etimation-webcam](notebooks/402-pose-estimation-webcam/) | Human pose estimation with a webcam or video file | <img src="https://user-images.githubusercontent.com/4547501/134550328-a5c99d22-ae60-4281-8120-a8f06a17b96a.png" width=225> |
```

ryanloney (changes requested):
The notebook works well for me on Windows. I requested changes to some of the text (just copy edits). I suggest in the future that we show how to download the video file used for the example, directly in the notebook. It's fine to wait until after you merge this, but it would be nice to have .mp4 that is downloaded automatically and run in the final cell. 

image

raymondlo84:
Screen Shot 2021-09-23 at 10 00 34 PM
Pretty cool, it seems to run on the Apple M1, but pretty slow :(

raymondlo84:
scrnli_9_23_2021_10-03-41 PM
and I got 2x the frame rate if I use FP32... Somehow the optimization got de-optimized on the M1 :D!!

adrianboguszewski:
Added video example

helena-intel:
@adrianboguszewski Thank you for the updates! The issue I ran into on Linux was caused by an unclean environment - with a clean env it works fine. I still see heavy flickering sadly, but the popup works. I consider that a workaround for now - ideally we would not need this. The OpenCV method as is also means that this notebook doesn't work when run remotely. A sentence in the Markdown above the video file and/or at the top of the notebook that explains (@ryanloney  can edit) "To use this notebook with a webcam, you need to run the notebook on a computer with a webcam. If you run the notebook on a server, the webcam will not work. You can still do inference on a video."

Can you change the popup video so that the text on the bottom is fully visible (or not there)? This is Ubuntu 20.04 on a clean env for me. Also the FPS was very fast for me, so that was invisible to distinguish the numbers. Could you update the FPS maybe every two seconds (not sure what would be the best number - please try a few options and see what works best!)
image

I am impressed by the frame from the video - I know these sample videos are not the most interesting but this one shows the result well. We can always use a more fun image/video without potential legal issues later.
The video at the end ran rather slowly for me (TGL, Ubuntu 20.04, CPU). Not in terms of FPS, but it looked like a slow motion video. How does it run for you? Live inference can be tricky - can you check the fps on this video and maybe reduce it to 15FPS for the demo? I do that in the monodepth notebook (as an option) with ADVANCE_FRAMES: https://github.com/openvinotoolkit/openvino_notebooks/blob/main/notebooks/201-vision-monodepth/201-vision-monodepth.ipynb I would also start a bit later in the video, when the people start coming in (add a skip_seconds or skip_frames parameter for example). It is nice if this notebook immediately shows results.

adrianboguszewski:
@helena-intel Unfortunately, I cannot do anything with the popup itself. Any problem with the window (e.g. not fully visible text at the bottom) is related to OpenCV compilation or your environment. I can only change the content you can see inside the window. 

I haven't seen more slow motion at the end. However, there is a little slow motion all time, but it comes from whole processing (not only inference). To have a better experience, we could use asynchronous processing. This is some complicated, so I propose creating another notebook to teach users async processing in OpenVINO.

I prefer not to add the next parameter. I believe every demo should be simple enough. skip_seconds or skip_frames would be necessary only in that particular video and not necessary in the case of a webcam.

Yeah, I think I can better calculate FPS. I'm going to improve this.

helena-intel:
> @helena-intel Unfortunately, I cannot do anything with the popup itself. Any problem with the window (e.g. not fully visible text at the bottom) is related to OpenCV compilation or your environment. I can only change the content you can see inside the window

It seems like you can remove the toolbar by using a `namedWindow` with specific flags: https://github.com/opencv/opencv-python/issues/163

> I haven't seen more slow motion at the end. However, there is a little slow motion all time, but it comes from whole processing (not only inference). To have a better experience, we could use asynchronous processing. This is some complicated, so I propose creating another notebook to teach users async processing in OpenVINO.

I understand why the slowdown is there, but it is not acceptable, and it can at least be partly fixed by changing the frame rate as I suggested. You process the video frame-by-frame, if the input video is 60fps but the total inference speed (including processing) is is less than that, you will see slowdown - possibly a lot. Using a 30FPS or 15FPS video would solve this because now the processing can keep up with the frame rate. It would still not be a full solution though, and neither would async. Unless the total processing+inference+display speed is the same as the fps of the video, you will get slowdowns or speedups, and that looks very strange in this kind of video, with people. So if you want to show live inference on a 15fps video you would need to add some code to make sure that the video shows at 15fps. 

> I prefer not to add the next parameter. I believe every demo should be simple enough. skip_seconds or skip_frames would be necessary only in that particular video and not necessary in the case of a webcam.

I think it is useful for many videos where you first have a bit of a boring intro before people, say, start dancing. And for this kind of demo you want to run the cell and see cool results, not stare at the screen for quite a few seconds thinking that it is just a still image of a store. But if you don't want to add it, please use a different video where you see the pose recognition immediately.

helena-intel:
https://user-images.githubusercontent.com/77325899/134815308-8126d0f6-e3a0-428e-99e6-684de3458f8a.mp4
https://user-images.githubusercontent.com/77325899/134815309-47575cf1-ae5a-4a71-9fbb-14bd69a28b5d.mp4

This shows the video output on my TGL laptop on Windows, with "better battery" power mode and "best performance" power mode.

ryanloney:
I tried running inside our notebooks container (built from the Dockerfile in main repo) and was not able to use the popout. This is on a laptop running Fedora 34 but using the UBI container. I'm ok if it does not work, but the only way to see the video was with flickering. 
The error: 
```
error: OpenCV(4.5.3) /tmp/pip-req-build-f51eratu/opencv/modules/highgui/src/window.cpp:1274: error: (-2:Unspecified error) The function is not implemented. Rebuild the library with Windows, GTK+ 2.x or Cocoa support. If you are on Ubuntu or Debian, install libgtk2.0-dev and pkg-config, then re-run cmake or configure script in function 'cvShowImage'
```

Seems to be the same as Helena saw before. 

Here is what the video playback looks like (flickering expected):
https://user-images.githubusercontent.com/15709723/134826210-95c59bca-b0a9-4c55-9fc1-2eae89036fc6.mp4

The webcam does not work when using the container. I'm sure there is a way to make it work in the future, since I've seen Google Colab demos that access the local webcam. Worth investigating for later. I'm not opposed to merging this for now, but I will leave it up to Helena and Raymond. At the least we should explain the limitations. 

Webcam error: 
```
Cannot open camera 0

[ WARN:0] global /tmp/pip-req-build-f51eratu/opencv/modules/videoio/src/cap_v4l.cpp (890) open VIDEOIO(V4L2:/dev/video0): can't open camera by index
```

raymondlo84:
@adrianboguszewski Let's catch up on this. Once we finished all these corner cases this will be a great starter for many other Notebooks in this repository.

adrianboguszewski:
I added my own video player to deal with FPS requirements. It's not an easy task, because there can be videos with lower or greater FPS than desired. Fortunately, the video player supports all situations. I hope it would be useful for all future live demos.

ryanloney:
It's looking better, I do not see flickering when using popout on Fedora now. I would really like the video in the store aisle not to flicker though, especially when embedded in the notebook. Anyone who uses Open Data Hub or OpenShift Data Science (where out notebooks are integrated) will have the annoying flicker. The other live inference videos that Helena embedded in other notebooks do not have this problem. Do you know why it happens in this notebook? (the video not the webcam) 

Screenshot of my test today: 
Screenshot from 2021-10-01 13-43-08

helena-intel:
Thanks @adrianboguszewski ! No more slowdown/speedup is great. The demo now works well for me. I found that if I use Edge on Linux (https://www.microsoftedgeinsider.com/en-us/download/?platform=linux) there is no flickering with the inline method.

notebooks/402-pose-estimation-webcam/402-pose-estimation.ipynb
```diff
@@ -0,0 +1,495 @@
...
+    "net = ie_core.read_network(model=model_path, weights=model_weights_path)\n",
+    "# load the model on the CPU (you can use GPU or MYRIAD as well)\n",
+    "exec_net = ie_core.load_network(net, \"CPU\")\n",
+    "\n",
+    "# get input and output names of nodes\n",
+    "input_key = list(exec_net.input_info)[0]\n",
+    "output_keys = list(exec_net.outputs.keys())\n",
+    "\n",
+    "# get input size\n",
+    "batch_size, channels, height, width = net.inputs[input_key].shape"
```
helena-intel, line 131: This shows a deprecation warning. Can you rewrite this with the new syntax that does not show this?

adrianboguszewski:
@ryanloney Flickering occurs for me only in Firefox. It is caused by the clear_output function, which is needed to clear the cell before showing a new image. So far I have no idea how to get around this. Today I tried some solutions but without success.

Helena uses a different type of processing. Firstly, she processes the video and after all displays it. Here I reuse the code for the webcam, so I process the frame and immediately display it.

PR #256, opened 2021-09-17
Add note about Windows and C++ to 302 notebook

helena-intel:
Temporary note that this notebook works on Linux and macOS and requires C++. I will add Windows support soon. @ryanloney please change the wording if you want.


ryanloney (approved):
LGTM

PR #258, opened 2021-09-18
Add cell for Windows compatibility to NNCF notebook

helena-intel:
Find location of cl.exe in MSVC installation to add to the PATH (for Windows). I don't think this is necessarily the best method - but it works for me locally and in the CI. We will add a Wiki page with guidance on installing C++ too that we can refer users to. 

Also minor changes to README, and add metadata cell to reduce test time by reducing dataset size in CI test.

### DIFF
--- a/notebooks/302-pytorch-quantization-aware-training/302-pytorch-quantization-aware-training.ipynb
+++ b/notebooks/302-pytorch-quantization-aware-training/302-pytorch-quantization-aware-training.ipynb
@@ -19,7 +19,7 @@
     "\n",
     "We selected the ResNet-18 model with Tiny ImageNet-200 dataset. ResNet-18 is the version of ResNet models that contains the fewest layers (18) and Tiny ImageNet-200 is a subset of the larger ImageNet dataset that should be downloaded in the notebook. Using the smaller model and dataset will speed up training and download time. To see other ResNet models, visit [PyTorch hub](https://pytorch.org/hub/pytorch_vision_resnet/).\n",
     "\n",
-    "> NOTE: _This notebook currently works on Linux and macOS and requires C++_"
+    "> NOTE: _This notebook requires a C++ compiler._"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


notebooks/302-pytorch-quantization-aware-training/302-pytorch-quantization-aware-training.ipynb
```diff
@@ -28,9 +28,42 @@
...
+    "# required C++ tools. This code assumes that Visual Studio 2019 is installed in the default \n",
+    "# directory. If you have a different C++ compiler, please add the correct path to os.environ[\"PATH\"] \n",
+    "# directly.\n",
+    "\n",
+    "import sys\n",
+    "import subprocess\n",
+    "if sys.platform == \"win32\" and subprocess.run([\"where\", \"cl\"]).returncode == 1:\n",
+    "    from pathlib import Path\n",
+    "    import os\n",
+    "    VS_INSTALL_DIR = r\"C:/Program Files (x86)/Microsoft Visual Studio\"\n",
```
ljaljushkin, line 55: Have you considered [vswhere](https://docs.microsoft.com/en-us/visualstudio/install/tools-for-managing-visual-studio-instances?view=vs-2019) tool instead? Potentially, it can be more general.


> vswhere.exe` is automatically included in Visual Studio 2017

notebooks/302-pytorch-quantization-aware-training/302-pytorch-quantization-aware-training.ipynb
```diff
@@ -376,7 +409,12 @@
      "base_uri": "https://localhost:8080/"
     },
     "id": "avCsioUYIaL7",
-    "outputId": "183bdbb6-4016-463c-8d76-636a6b3a9778"
+    "outputId": "183bdbb6-4016-463c-8d76-636a6b3a9778",
+    "tags": [],
+    "test_replace": {
+     "80000, 20000": "300, 100",
```
ljaljushkin, line 415: 👍

notebooks/302-pytorch-quantization-aware-training/302-pytorch-quantization-aware-training.ipynb
helena-intel: I did! I came across it thinking surely there must be a better way. But I found it not very intuitive, it is only included from VS 2017+ (I have no idea how common VS 2015 still is - that may not be an issue) and you would still need to assume the installation in C:\Program files (x86). Or you would have to download the vswhere executable in the notebook, but that seemed excessive and potentially triggering security flags. If you have a simple suggestion for how to use `vswhere` here I would be happy to use that. 

I also think there is a smarter way even with path searching, but I did not want to make the code in this cell even longer just to do this check for Windows users. Improvements welcome though!

ljaljushkin (approved):

helena-intel:
I also made a small change to the notebook headers: for the integration in documentation, there can only be one top level (H1) header in a notebook. I moved all headers apart from the title one level done (H1->H2, H2->H3 etc.) so the structure remains the same.

ryanloney (approved):
This has issues on my Windows machine, Helena is troubleshooting, but I don't want to hold up since it passes CI and works on other machines.

helena-intel:
For now I added a note that if importing fails on Windows, you can add the path to the directory that contains python3[6,7,8].lib to `os.environ["LIB"]`. This is not always required and I do not yet know why it fails on some computers but not everywhere. I'll look into that.

PR #260, opened 2021-09-20
Minor editorial update of NNCF tutorial

MaximProshin:


helena-intel (approved):
Thanks @MaximProshin ! I also fixed this in https://github.com/openvinotoolkit/openvino_notebooks/pull/258 but that's still under review. I'll merge this now.

PR #235, opened 2021-09-06
Fix download file function and modify progress bar

Debskij:
Fix bug reported in [issue 226](https://github.com/openvinotoolkit/openvino_notebooks/issues/226) and attempt to solve [issue 230](https://github.com/openvinotoolkit/openvino_notebooks/issues/230). 230 isn't yet ready for closing and needs further testing for TQDM memory performance.

PR #261, opened 2021-09-20
Enable progress bar for download of MRPC.zip in 105

helena-intel:
Progress bar was temporarily removed from 105 because it did not always work well on macOS. A new progress bar was added in #235 that should fix this issue, so the progress bar can be enabled again.


ryanloney (approved):
Works for me on macOS
Screen Shot 2021-09-21 at 22 50 13

PR #263, opened 2021-09-22
Add Python library path to LIB for Windows

helena-intel:
* Add Python library path to `os.environ["LIB"]`. 
* Ran Jupyterlab Code Formatter (out of habit...), which changed some single quotes to double quotes and changed some newlines. These are just formatting changes, the only actual code changed is around line 70:
image
* Changed the Jupyter kernel to openvino_env.

I tested on a fresh install of Python in Windows Sandbox that python37.lib exists in this path.  Also tested on a Windows install that used to raise an error about not finding python37.lib that this no longer occurs after adding these lines. More testing would be welcome.


ljaljushkin (approved):

ryanloney (approved):
Tested on Windows. It works for me now!

PR #267, opened 2021-09-23
Dockerfile - optional --build-arg for GPU dependenices

ryanloney:
Installs dependencies for integrated GPU. I will add instructions to the wiki after this is merged, but basically you just need to run this to build: 
```
docker build -f ./Dockerfile_gpu . -t openvino_notebooks
```

And this to run the image:
```
docker run -it --device /dev/dri -p 8888:8888 --shm-size 8G openvino_notebooks
```

I'm showing a demo on Tuesday on Tiger Lake laptop with Fedora. I have already tested and confirmed that this works. It uses the same approach for installing iGPU dependencies as OpenVINO Runtime and Model Server containers that we publish on Red Hat registry and Docker Hub.


helena-intel:
Thanks @ryanloney ! Is it an option to just put this into our regular Dockerfile? The image is already large so it should not add that much (I can test), and that would avoid code duplication and maintenance effort.

ryanloney:
@helena-intel yes, but we should not include this in default build, not everyone will need it, especially for our integration with ODH/ODS, which runs primarily on Xeon VMs in the cloud. 

I suggest we use a "build-arg" for GPU. Something like  `--build-arg GPU=True` for our docker build command?

It will take me some time to figure this out. If you want to try, or have a suggestion for me to try please share it.

ryanloney:
@helena-intel I updated the main Dockerfile so it installs the GPU dependencies when using the `--build-arg` flag to set `GPU=True`. 

To build: 
```
docker build --build-arg GPU=True . -t openvino_notebooks
```

To run:
```
docker run -it --device /dev/dri -p 8888:8888 --shm-size 8G openvino_notebooks
```

helena-intel:
Thanks @ryanloney! I really like having it in the same file. 
It does not work out of the box on my TGL laptop with Ubuntu 20.04. I get 

```
[Warning] One or more build-args [GPU] were not consumed
```
I ran the 002 notebook in Docker to check if GPU was working anyway, but it was not. 

I locally installed the GPU with `apt install intel-opencl-icd` and it works fine then:
image

If it works for you and you would like to merge now I can do that and look into this later.

ryanloney:
@helena-intel I just pushed an update. `ARG GPU=False` needs to be set by default. I left it blank before. Can you try again?

adrianboguszewski:
@ryanloney It doesn't work for me with Ubuntu as well.
Screenshot from 2021-10-08 12-36-31

raymondlo84:
Let's do a clean up and see if this pull request is still relevant?

alberto139:
Hello,

I know this issue has been closed for a while, but I wanted to check in if there's a way to use openVINO with an iGPU through a docker container.

I've also attempted to build on Ubuntu 20.04 with the suggested Dockerfile with no luck.

adrianboguszewski:
@alberto139 see if the instructions from #793 can help you

ryanloney:
@dtrawins do you have any advice on how to get OCL working in the container?

PR #268, opened 2021-09-24
Add in jupyterlab-git plugin

Xaenalt:
Should check that this doesn't break dependencies if you've got different versions than what thoth-station uses, but other than that this should be all you need I think

### DIFF
--- a/.docker/Pipfile
+++ b/.docker/Pipfile
@@ -15,6 +15,7 @@ jedi = "==0.17.2"
 jupyter-nbrequirements = "*"
 jupyterhub = "==1.1.0"
 jupyterlab = "*"
+jupyterlab-git = "==0.30"
 matplotlib = "<3.4"
 nbval = "*"
 nncf = {extras = ["torch"], version = "*"}
--- a/.docker/Pipfile.lock
+++ b/.docker/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "62b6f0d2e17a512794caffa2adf4e9873c5586e3eabe462634c2efb983732e21"
+            "sha256": "0b996db02b65d118d83e68b1705e49399fb9c618b63be21ca77b146c22f80a0a"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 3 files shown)


ryanloney:
@Xaenalt the build failed, we will need to look into it more but if you want to see what happened: https://github.com/openvinotoolkit/openvino_notebooks/pull/268/checks

Xaenalt:
Huh, weird, something about supervisord there, checking my other build to be sure

bosbes:
Thanks for the contribution! I don't know if this is what causes the test to fail, but I notice in https://github.com/openvinotoolkit/openvino_notebooks/pull/268/files that Pipfile is changed, but not Pipfile.lock - that will cause issues.

helena-intel:
Thanks for the contribution! I don't know if this causes the test to fail but I notice in https://github.com/openvinotoolkit/openvino_notebooks/pull/268/files that Pipfile is changed and Pipfile.lock is not. That will cause problems.

Xaenalt:
Ok, figured out the issue, Pipfile.lock has a checksum, and it's being difficult on my system, should have the PR cleaned up soonish xD

helena-intel:
> Ok, figured out the issue, Pipfile.lock has a checksum, and it's being difficult on my system, should have the PR cleaned up soonish xD

Aah! Thanks @Xaenalt that is useful to know. If it keeps being difficult I can of course also add this change next week. I imagine it must be annoying to make a tiny quick PR and then struggle with CI failures.

Xaenalt:
I have no idea how it computes the hash, but ideally this one should do it

Xaenalt:
If this doesn't do it, then I officially have no clue how they compute that hash xD

Xaenalt:
Alright, moment of truth

helena-intel:
> Alright, moment of truth

Totally worked it seems! Thank you for your persistence! So do I understand correctly that you manually had to update this hash? I'm now worried that I will have this same issue next time I want to update this file :-) 

Also wondering if you need  this version of jupyterlab-git, or if we can upgrade it in a future build if necessary for compatibility/security reasons.

helena-intel (approved):
Thank you!

PR #269, opened 2021-09-24
Add model caching to 002 notebook

helena-intel:
Add model caching cell to API tutorial. 

I fixed it to GPU device for now because caching is not available for CPU and we have not covered other devices anywhere yet. I can expand this later.

There are a few style updates in the diff too (from black) - the only content I added is the final section and an entry to the table of contents.


ryanloney:
This is incredible! 

On Tiger Lake laptop with Fedora 34:
image

16s GPU warmup down to 120ms with model caching. What an amazing feature!

ryanloney (approved):
I tried on Linux and Windows, similar results on 8th and 11th Gen Core i7 processors. LOVE THIS ADDITION! Thank you!

helena-intel:
It is amazing indeed! It makes working with GPU so much better. 20 seconds wait every time adds up. We can add this to some notebooks where  GPU performance matters too, but need to do that without introducing too much boilerplate.

PR #241, opened 2021-09-08
Speed up CI tests by patching notebooks

helena-intel:
Add a script to the nbval CI to patch notebooks before the CI is run. The patch_notebooks script replaces values in the notebook, based on metadata in the notebook. For example: nr of epochs is changed from 15 to 1 in  301-tensorflow-training-openvino-pot.ipynb by adding  `{"test_replace": {"epochs = 15": "epochs = 1"}` to the cell  metadata of the cell that contains `epochs = 15`. (Test notebooks are written to test_notebookname.ipynb so the patch_notebooks script does not overwrite existing notebooks - this is useful if you want to run this script locally too).

This implementation is not ideal - but there aren't really any ideal implementations. This works and makes the tests run a lot faster. I added the 301 notebooks back to the CI in this PR (will add 208 later). 

Currently implemented replacements:

- Number of frames in superresolution

> 202-vision-superresolution-video.ipynb: NUM_FRAMES = 100 -> NUM_FRAMES = 3

- Smaller dataset size and statistics subset size for POT notebooks

> 105-language-quantize-bert.ipynb: return len(self.dataset) -> return 30
105-language-quantize-bert.ipynb: "stat_subset_size": 250, -> "stat_subset_size": 25,
301-tensorflow-training-openvino-pot.ipynb: "stat_subset_size": 1000, -> "stat_subset_size": 25,
301-tensorflow-training-openvino-pot.ipynb: return len(self.dataset) -> return 30

- Run benchmark_app for 3 or 5 seconds instead of 15 or 60 (could standardize to 3 everywhere)

> 105-language-quantize-bert.ipynb: benchmark_app -> benchmark-app -t 5
105-language-quantize-bert.ipynb: benchmark_app -> benchmark-app -t 5
104-model-tools.ipynb: -t 15 -> -t 3
104-model-tools.ipynb: seconds=15 -> seconds=3
104-model-tools.ipynb: seconds=15 -> seconds=3
104-model-tools.ipynb: seconds=15 -> seconds=3
104-model-tools.ipynb: seconds=15 -> seconds=3
301-tensorflow-training-openvino.ipynb: epochs = 15 -> epochs = 1
301-tensorflow-training-openvino-pot.ipynb: 301-tensorflow-training-openvino.ipynb -> test_301-tensorflow-training-openvino.ipynb
301-tensorflow-training-openvino-pot.ipynb: -t 15 -> -t 3
301-tensorflow-training-openvino-pot.ipynb: -t 15 -> -t 3
[...]

PR #276, opened 2021-09-29
Add NNCF tutorial based on TensorFlow framework with Keras API

negvet:


negvet:
Hi! 
I do `import tensorflow_datasets as tfds` in the first cell. 

But CI fails on the second cell execution with 
```
NameError                                 Traceback (most recent call last)
<ipython-input-1-9e939355bc76> in <module>
----> 1 datasets, datasets_info = tfds.load('imagenette/160px', as_supervised=True, with_info=True)
NameError: name 'tfds' is not defined
```
I am not sure if I know how to fix it. Locally it is not reproducible.

Please help!

helena-intel:
Thanks for this notebook! I think the rerunfailures plugin for pytest causes this error message that is not actually causing the test to fail - possibly this plugin doesn't work well with Jupyter notebook. I removed the plugin in https://github.com/openvinotoolkit/openvino_notebooks/pull/277 and will merge once the tests pass. Then you'll hopefully see what actually goes wrong. The test doesn't fail locally for me either in that cell - so it may be something with the CI.

To make the CI run not take too long, you can patch the notebook, for example to run for 1 epoch instead of 4, or to use a small subset of the dataset for the test. See https://github.com/openvinotoolkit/openvino_notebooks/pull/241

When I run the notebook locally (Ubuntu 20.04, Python 3.8), the kernel dies on this cell:
image
Do you recognize that?

negvet:
Hi @helena-intel ,

I am still experiencing some issues with the CI. Could you please help me to solve them? 
Below are some questions that could help me to figure it out.

1. All CI checks are run on machines with CPU only? GPU not used for validation? If yes, do we have some options to disable CPU CI builds, since training on CPU could take too long for the computer vision problem? We also could provide a NOTIFICATION on the top in the notebook that to run you need to have GPU available. Currently, I use a dataset that is 94Mib - not that big. I also would like to use the ResNet-50 model.

2. Please tell me what is `Timeout of 2000 seconds exceeded executing cell` I am getting it on the CI. Is it when the cell is not responding (frozen) for >2000 secs? Or is it when the cell is executing for >2000 secs? I am asking because if the cell is training for >2000 secs it is actually not frozen. Any options to increase timeout?

3. Unfortunately I am not able to reproduce any freezing of the notebook. Could you please run the current version of the notebook locally? If it still fails please send the detailed instructions (with the environment) to reproduce it (or better docker file).

helena-intel:
Thanks for the updates @negvet ! I now no longer experience the crash that I saw before (on the same laptop), that is great.

We do not expect that users have a GPU (and I do not have one myself on my development laptop). I do recognize that for NNCF this could be helpful. Is using a smaller model an option for this notebook? I am training on CPU now - it is slow, but it is what it is.  Hopefully we can soon use TensorFlow 2.5 with OneDNN speedup in openvino-dev, that should also help. Is training for 1 epoch enough to show reasonably good results? There could then be instructions for users who do have a GPU to train for more epochs. 

For the CI, you can patch the notebooks by editing metadata. For the 302 notebook I reduced the dataset size that way - that helped a lot. https://github.com/openvinotoolkit/openvino_notebooks/wiki/Notebooks-Development---CI-Test-Speedup Can you try that? That should fix the 2000 seconds error as well.

negvet:
Sorry, messed up with git. will open a new PR

Issue #273, opened 2021-09-28
Git pull causes pod to crash loop on new push to main


Xaenalt:
**Describe the bug**
In cases where the user has modified or done anything to the default notebooks, a new commit to main on github can cause pods to crash loop

 **Expected behavior**
User modifications should not impact functionality of the notebook image

**Screenshots**
![image](https://user-images.githubusercontent.com/8479010/135145973-adf213e0-a5ca-4988-9b09-2df60429cb42.png)

 **Installation instructions (Please mark the checkbox)**
[X] I followed the installation guide at https://github.com/openvinotoolkit/openvino_notebooks#-installation-guide to install the notebooks. 

 
**Additional context**
It stems from this addition

https://github.com/openvinotoolkit/openvino_notebooks/blob/16de32f51f701fa5d49301333582054972cb2c4c/.docker/builder/run#L49

The point of the preload environment variable is to pre-load content for demos/labs/etc, however adding the git pull causes a problem when the user attempts to do any modifications.

To reproduce:
1. Checkout an older version of the repo, eg: `git checkout HEAD^`
2. Replace `main` branch with it, so that main now points to an older commit. Eg: `git branch -D main; git switch -c main` (this simulates the effect of a new commit being pushed to main upstream)
3. Make any change to the code, editing any line in any notebook
4. Restart the container
The pod will now crash loop, and will continue to do so until the user either manually intervenes outside of the OpenVINO image or deletes their PVC

I'd propose a solution would be either to version the selected notebooks into the image, or use the preload variable as intended. Additionally, if it's absolutely necessary to have this functionality, the notebooks could get cloned/updated in a pristine condition somewhere else in the image, and then copied into the user's volume to avoid this type of conflict

I can draft a PR to address

helena-intel:
Thanks @Xaenalt ! If you want to make a PR that fixes this am happy with any solution that works for you.

Xaenalt:
So question, how critical of a feature is this git pull thing? I'm thinking just preloading the notebooks into the image, and copying them into place would be better

Also, is there a plan for version tagging? If so that would be good to pin library versions for RHODS as well

helena-intel:
We currently have no plan for version tagging - we recommend everyone to always run the main version since that is up-to-date with bug fixes and improvements, and tested daily in the CI. Since we support 3 different Python versions and 5 operating systems and we want to have a simple requirements.txt for users, it is not possible to have releases that are known-to-work on all supported configurations. 

[...]

PR #279, opened 2021-10-01
Update to remove the git pull since it can cause crashes

Xaenalt:
Resolves https://github.com/openvinotoolkit/openvino_notebooks/issues/273
This handles the issue with git pull crashing a container if there's a merge conflict

It instead places the notebooks in /opt/app-root/notebooks, and copies them into ~/openvino_notebooks at runtime if the directory doesn't already exist

This also has the benefit of locking the notebooks to the current version. Once a git push can trigger a rebuild, the loop will be closed on maintenance

### DIFF
--- a/.docker/builder/run
+++ b/.docker/builder/run
@@ -44,14 +44,15 @@ if [ -n "${JUPYTER_PRELOAD_REPOS}" ]; then
         fi
         echo "Checking if repository $repo exists locally"
         REPO_DIR=$(basename ${repo})
-        if [ -d "${REPO_DIR}" ]; then
-            pushd ${REPO_DIR}
-            GIT_SSL_NO_VERIFY=true git pull --ff-only
-            popd
-        else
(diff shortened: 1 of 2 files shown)


helena-intel:
Thanks @Xaenalt ! The PR is appreciated. I added a commit that should fix the failing docker_nbval test.

Xaenalt:
Oh, let me rearrange that real fast

Xaenalt:
That should cover it under the fix-permissions call

It looks like it might have clobbered the fix tests commit though

helena-intel:
Thanks @Xaenalt ! Github Actions is very slow today, but I'll keep an eye on the tests.

ryanloney (approved):

helena-intel:
I'll merge now and fix the test in a new PR. I tested locally that the image builds and runs. Thanks again @Xaenalt !

PR #280, opened 2021-10-01
Fix Docker CI test

helena-intel:
See #279 , this PR fixes test that failed after merging that.


helena-intel:
@Xaenalt FYI, I added a `RUN chown -R 1001:0 /opt/app-root/notebooks` line. Please let me know if you think it's not a good idea.

ryanloney (approved):

Xaenalt:
Cool, yeah I wasn't sure why that ended up happening, I had intended to move it there to the fix-permissions call would get it, but just adding the explicit chown works just as well

Issue #215, opened 2021-08-28
Pull Request 208 Revisited - Bugs & Improvements


raymondlo84:
Would like to track the bugs here again as we have a few openings. Thank you. 

https://github.com/openvinotoolkit/openvino_notebooks/pull/209

```
Hi, @Debskij

We should use the util functions Helena made on downloading files.

def download_file(
    url: PathLike, filename: PathLike = None, directory: PathLike = None, show_progress: bool = True
):
    """
    Download a file from a url and save it to the local filesystem. The file is saved to the
    current directory by default, or to `directory` if specified. If a filename is not given,
    the filename of the URL will be used.

    :param url: URL that points to the file to download
    :param filename: Name of the local file to save. Should point to the name of the file only,
                     not the full path. If None the filename from the url will be used
    :param directory: Directory to save the file to. Will be created if it doesn't exist
                      If None the file will be saved to the current working directory
    :param show_progress: If True, show an IPython ProgressBar.
Also, we shall print out the command and the model downloader is not showing any progress bar. I thought it was crashed and we shall have some ways to show progress there.

    # Run commands, first download model than convert it to inferable 
    !! $download_command
    print('Models downloded')

@raymondlo84
 
Member
raymondlo84 commented 4 days ago
plt.imshow(convert_result_to_image(image, resized_image, boxes_with_annotations, conf_labels=True))

We may want to size it up so the result is easier to read. The text at the end is very hard to read.

@ryanloney
 
Member
ryanloney commented 4 days ago
plt.imshow(convert_result_to_image(image, resized_image, boxes_with_annotations, conf_labels=True))

We may want to size it up so the result is easier to read. The text at the end is very hard to read.

@Debskij - can you fix this in a new PR? ^^

```

PR #281, opened 2021-10-03
208-add spinner, change download dir, visualize boxes

### LINKED ISSUE #234: Progress bar could be useful in omz_downloader
There is no progress bar during model downloading in `208-optical-character-recognition` (third code cell). Currently, there is no information when downloading ends. 

Such progress bar could be useful when someone (like me) has a slow internet connection.

helena-intel:
- Add spinner to show download progress (with yaspin)
- Change OMZ download directories
- Add visualization of OCR on extracted boxes
- Add ~/open_model_zoo_cache to Github Actions CI and add 208 back to CI
- Use load_image to load image from url or file
- Add E703 (semicolon at end of line) to nbqa exclude (semicolon prevents displaying useless return values)

Closes #234 and addresses issues mentioned in #215 

OMZ models are downloaded to standard directories ~/open_model_zoo_models with cache dir ~/open_model_zoo_cache to allow users to use Open Model Zoo's `--all` or `--list` command to download several models at once when they have fast internet and then run the notebooks that use this later. We could explain this in the README/start of the notebooks that use OMZ models. 

To facilitate downloading the model from the Jupyter interface I copy the model in the model directory. It seems better to have file redudancy than download redundancy. I tried to use `os.link` instead of copy. That worked great locally, but gave an error in the CI because on Windows a hard link cannot span different devices. 

Screenshot of added output (below original output)
fe07e47ffaa9d899d6b2568dc2721293

### DIFF
--- a/.github/workflows/codecheck.yml
+++ b/.github/workflows/codecheck.yml
@@ -41,7 +41,7 @@ jobs:
         python -m ipykernel install --user --name openvino_env
     - name: Flake8
       run: |
-        nbqa flake8 --ignore=E124,E203,E266,E402,E501,F821,W503,W291,W293 --nbqa-exclude="(301.*)|(302.*)|(208.*)" notebooks
+        nbqa flake8 --ignore=E124,E203,E266,E402,E501,E703,F821,W503,W291,W293 --nbqa-exclude="(301.*)|(302.*)" notebooks
     - name: Check READMEs
       run: |
         python -m pytest .ci/test_notebooks.py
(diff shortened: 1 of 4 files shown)


ryanloney:
I like these enhancements. Here is the output of an image that I used to test. It is really nice to see the detected boxes and recognized text next to each other. 

Screenshot from 2021-10-03 17-36-57

Screenshot from 2021-10-03 17-37-04

ryanloney (approved):
Thank you for updating some of the descriptions and adjusting the step to move OMZ models. The larger output with detected boxes is also much easier to read. Thanks for setting such a high bar for the notebooks!

PR #283, opened 2021-10-05
Add requirements check CI test

helena-intel:
We have three places where requirements are defined: requirements.txt, .binder/requirements.txt and .docker/Pipfile.

This PR changes the .binder/requirements.txt to just `-r ../requirements.txt` so everything from requirements.txt is automatically in the Binder requirements. This is not strictly necessary - there are some notebooks that do not run on Binder, and Binder automatically includes Jupyter. But the frameworks are our largest dependencies, and Binder images are build only once (after repository change) so the overhead is very small, and the benefit of not having to keep two files up-to-date is large.

It is not as easy to keep .docker/Pipfile and requirements.txt in sync. This PR adds a CI test that checks that all requirements from requirements.txt exist in .docker/Pipfile. It also tests this for .binder/requirements.txt - which is not necessary for the current version, but is useful in case someone changes the Binder requirements or setup later.

I tested manually that Binder still works, and that there are no conflicting requirements there. 

Other small changes:
- Add yaspin to Docker image
- Add all requirements with version pinnings from requirements.txt to Pipfile
- Remove tqdm and requests from requirements.txt (already in openvino-dev)


ryanloney:
Is there a way to automate the building of Pipfile.lock using GitHub Actions? Not something we need to do right away, but long term it would simplify the updates when updated versions of dependencies are needed in the Dockerfile.

helena-intel:
> Is there a way to automate the building of Pipfile.lock using GitHub Actions? Not something we need to do right away, but long term it would simplify the updates when updated versions of dependencies are needed in the Dockerfile.

That would be ideal, but when I looked into automating things from a PR from a fork I ran into annoying permission issues. We could have it on merge, but we want to test the notebooks before merge.

ryanloney (approved):

PR #46, opened 2021-04-26
Add facies segmentation demo

alexeyhorkin:
I want to add facies segmentation demo =)

PR #295, opened 2021-10-15
Remove openvino-extensions package

dkurt:
`openvino-extensions` is a custom package designed for https://github.com/openvinotoolkit/openvino_notebooks/pull/46/. As the demo reverted, do not install it.


ryanloney:
Thanks @dkurt. We hope to host the facies notebook when the model weights have a reliable storage location, like Amazon S3. For now we can remove openvino-extensions dependency though.

dkurt:
Hi! I have issues with using `pipenv` (it hangs without response). So if you are able, feel free to reopen a PR.

helena-intel:
I locked the Pipfile, but it seems there is an unrelated upstream issue with the Docker image now. I'll keep an eye on it. Once the tests pass I'll merge the PR (and if there is still an issue I'll fix it).

I'm sad that we cannot have the facies notebook at the moment. I still hope we can add it soon.

CONTRIBUTING.md
```diff
@@ -155,15 +156,30 @@ To maintain consistency between notebooks, please follow the directory structure
...
 In case of output provided by Notebook please create folder ```output``` on the same level as readme file.
 
+### Notebook utils
+
+The _notebook_utils.py_ file in the _notebooks/utils_ directory contains utility functions and classes that can be reused across
+notebooks. It contains a `download_file()` function that optionally shows a progress bar, and a standard way to convert
+segmentation maps to images and display them. The Python file is generated from _notebook_utils.ipynb_ notebook in the same directory.
+If you want to add a function or class to _notebook_utils.py_, please add it to the notebook, and generate the
+Python file with `jupyter nbconvert notebook_utils.ipynb --TagRemovePreprocessor.remove_cell_tags=hide --to script`
+Add a "hide" tag to any demo cells (from the right side gear sidebar) to prevent these cells from being added to the script.
```
dkurt, line 166: @helena-intel, Seems like there are unrelated changes

helena-intel: @dkurt I rebased to main, the changes should all be from that. I'll doublecheck before merging.

PR #297, opened 2021-10-17
Fix header on 301

Debskij:
Missing level 1 header was breaking documentation


helena-intel:
Thanks @Debskij ! We should have a CI test that checks this.

PR #316, opened 2021-10-26
107 Quantize omz model ssd mobilenet v1 coco

trexdog:
Add JNB 107 to Optimize and quantize the Open Model Zoo ssd_mobilenet_v1_coco model


trexdog:
Please note that as of now JNB 107-quantize-omz-model-ssd_mobilenet_v1_coco is complete here as a standalone, however it will no longer be updated because it is being merged with JNB 108-quantize-omz-model-yolo-v4-tf on a new branch which will contain a new merged JNB 107-name-to-be-decided-soon that can run using either of the two models ssd_mobilenet_v1_coco and yolo-v4-tf.

PR #317, opened 2021-10-26
108 Quantize omz model yolo v4 tf

trexdog:
Add JNB 108 to Optimize and quantize the Open Model Zoo yolo-v4-tf model


trexdog:
Please note that as of now JNB 108-quantize-omz-model-yolo-v4-tf is complete here as a standalone, however it will no longer be updated because it is being be merged into the quantize-omz-model-ssd_mobilenet_v1_coco branch which will contain a new merged JNB that can run using either of the two models ssd_mobilenet_v1_coco and yolo-v4-tf.

PR #329, opened 2021-11-11
Add JNB 107 to optimize and quantize the OMZ object detection models yolo-v4-tf and ssd_mobilenet_v1_coco 

trexdog:
JNB 107-quantize-omz-object-detection-models which is a merge of and supersedes the previous 107-quantize-omz-model-ssd_mobilenet_v1_coco and 108-quantize-omz-model-yolo-v4-tf JNBs.


trexdog:
Note that runs on Windows are failing due to timeouts due to omz_quantizer taking more than 2000 seconds.  It is expected to be slow for the yolo-v4-tf model which is the default.  Option may be to make the ssd_mobilenet_v1_coco model the default since it, at least for build purposes, will run faster.

helena-intel:
Thanks @trexdog ! I would like the SSD model to be the default. I'm going to look into making the test run faster for this notebook, but this would help for now, and It is nice for users too if the first time they run the notebook it runs faster.

ryanloney:
@trexdog is there something wrong with the labels? The airplane is detected as a bus:
Screenshot from 2021-11-11 19-49-20

And the accuracy seems really poor:
Screenshot from 2021-11-11 19-51-00

trexdog:
The detection for the ssd_mobilenet_v1_coco model is poor, look at the OMZ site which states "coco_precision 23.3212%".  
Try the other images, some are better than others.

trexdog:
I mixed up the error with what was common, there was actually an off-by-one error from merging the two JNBs which each have different class offsets.  Many of the plane pictures do get identified as buses, but the one chosen does get identified correctly.  The plane should now be labeled "airplan" due to the text in the OMZ labels file.  The bear and elephants should be correct too.

PR #334, opened 2021-11-18
Add note about POT notebook to 301 training notebook

helena-intel:


MaximProshin (approved):

raymondlo84:
One thing that is a bit confusing is the use of saved model, vs frozen graphs, vs OpenVINO IR etc... It is a little bit all over the place as I read through the documentations.

Also, let's avoid putting, there will be error or warning etc, without providing instructions of how to fix it. For example, why is the saved models does not work? ...? 

I think it will be great if we can make these terms more consistent across. As sometimes we say SavedModel, saved model, and Frozen Graphs vs Frozen graph vs FrozenGraph vs Frozen Graph models vs Frozen Graph format... 

I made a quick touch up and let's keep making it better so the instructions are more clear. The code works however, that's the best news :)

helena-intel:
Thanks @raymondlo84 but it seems your comments and edits are for a different notebook than 301. Could you make a new PR with these changes?

raymondlo84 (approved):
<3 Thanks for the updates.

PR #337, opened 2021-11-22
Add cache files to all workflows

helena-intel:
Add .env file to hold environment variables: PIP_CACHE_KEY and FILES_CACHE_KEY for caching across workflows, HUB_HOME to set PaddleHub HUB_HOME to /tmp/paddlehub to allow caching


helena-intel:
image

This is how you see that the cache works: the "Cache OpenVINO dependencies" step is skipped, and the "Install dependencies" step shows that openvino-dev is collected from the cache.

PR #339, opened 2021-11-23
Add third-party-programs.txt and clarify dependencies in requirements.txt

helena-intel:
* Add third-party-programs.txt file with license information from third party components which sources we use. 
* Clarify in requirements.txt which entries are dependencies-of-dependencies and are only in the requirements.txt to pin to a specific version, to avoid vulnerabilities or issues.


raymondlo84 (approved):
Cool! Actually a good update to the requirement.txt

ryanloney (approved):

PR #352, opened 2021-12-11
Pin paddlenlp in requirements.txt

helena-intel:
paddlenlp pinned to 2.0.8 to prevent ImportError that causes 103 to fail

image


helena-intel:
Removed numba pin from this PR because it caused a CI test to fail. It is not actually necessary to pin numba. If you now install all the requirements, numba 0.53.1 is installed, which is not vulnerable. 

nbval CI passes again with paddlenlp pin, merging now.

PR #363, opened 2021-12-22
Add `net.serialize()` to 002

helena-intel:


ryanloney (approved):
I know it doesn't get much easier than `mo --input [model].onnx` but nice that we can also read in an ONNX and export an IR from OpenVINO API. Thanks for adding this to the tutorial.

PR #365, opened 2021-12-23
213 Interactive question answering

adrianboguszewski:


helena-intel:
See the BERT demo I made while ago (did not make it to the repo): https://github.com/helena-intel/openvino_notebook_demos/blob/main/bert-qa/bert_question_answering_demo.ipynb

helena-intel:
@adrianboguszewski would it be possible to connect this to the [BERT quantization notebook](https://github.com/openvinotoolkit/openvino_notebooks/tree/main/notebooks/105-language-quantize-bert)? We really miss a demo there.

adrianboguszewski:
@helena-intel I'd prefer not to mix this content with 105. I think we should have "nice and easy" interactive notebooks in 4xx branch to quickly demonstrate possibilities and to make a "wow" effect.
In my opinion, 403 should be the playground and 105 should teach people how to quantize. But of course, we could add some simple demo in 105, but not a looped interactive one.

helena-intel:
Thanks @adrianboguszewski I did not mean to add bells and whistles to 105, and I agree a separate demo is nice. It would be nice if they were connected, and for example use the same model, so that we can have a simple inference demo in 105 indeed, and that we can refer people from this demo notebook to 105 that teaches how the model was created.

Should this demo notebook be in the 200 series? 400 was created because for best results these notebooks need a webcam, which not everybody has, and which does not currently work on cloud servers.

ryanloney:
I agree this should be in the 200 series. 
Similar to the other interactive notebooks, we need to provide a sample input ("When was OpenVINO created?" or something) that shows an answer when a user runs all cells. Then they can type in questions. 

We should also re-use some of the code for 105, so it can answer a question or compare two sentences or something.

raymondlo84:
The performance seems a bit too low for the Intel wiki page... I'm getting 7s and the answer seems to be a bit all over the place...  Maybe we shall pick a slightly smaller page and so we can have reasonable performance?


Context: ['https://en.wikipedia.org/wiki/Silicon_Valley']
 who's in silicon valley
Question: who's in silicon valley
Answer: Silicon Valley has been home to the largest concentration of venture capital firms in the world
Time: 2.04s

helena-intel:
image
Results are not great indeed. If this model only works well with relative short websites, I agree with changing the URL. There should also be some info about the limitations of the model.

ryanloney (approved):

PR #378, opened 2022-01-18
Pin torchmetrics and set User-Agent in 301

helena-intel:
Torchmetrics was updated and the API has changed. Pin to known working version for now. 301 sometimes fails now with a message to set a user agent. I added that to the 301 notebook, and added a check in the POT notebook that only downloads if the image does not exist. We could use `download_file()` from notebook_utils here, but I did not want to change too much in this notebook now.


helena-intel:
CI passes, merging now.

PR #388, opened 2022-01-27
Add install_requirements_macos CI workflow

helena-intel:
We don't run macos in the nbval tests currently because there were often issues with notebooks that have many outputs (like 401/402). This test at least checks that the requirements install on macOS. It runs daily.

Also added macos to install_requirements.yml. That test does not run on a schedule or on changing requirements.txt, because nbval already does and that implicitly tests installing requirements. install_requirements.yml is used for manually testing that the requirements (still) install.

When the nbval tests are fixed for macos the install_requirements_macos.yml test can be deleted, everything is covered by nbval/install_requirements.yml then.

### DIFF
--- a/.github/workflows/install_requirements_macos.yml
+++ b/.github/workflows/install_requirements_macos.yml
@@ -1,20 +1,28 @@
-name: install_requirements
+# Test that requirements install on macOS 10.15
+# Unlike install_requirements.yml this workflow runs daily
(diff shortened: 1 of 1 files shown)


ryanloney (approved):

helena-intel:
Closing in favor of https://github.com/openvinotoolkit/openvino_notebooks/pull/396

PR #391, opened 2022-01-28
Update 210 README with new CT scan gif

helena-intel:
The current gif for 210 is for an older version of the notebook. This PR replaces the gif with the same gif that is in 110, which shows faster inference.


raymondlo84:
Would be cool to put that gif in a loop mode. https://www.veed.io/tools/loop-video/loop-gif

ryanloney (approved):

PR #393, opened 2022-01-31
Fixed video player test, added score to 213

adrianboguszewski:


helena-intel:
Thanks @adrianboguszewski , do you have any idea what causes the issue? Wondering if this is a workaround for the CI, but users might still be affected. When this fails, it is most often in notebook_utils, but not always, I also see it in 401, for example here: https://github.com/helena-intel/openvino_notebooks/runs/5004571900?check_suite_focus=true

raymondlo84:
Is that validated? Is this going to block Paula's merge too? Please make sure it's validated.

helena-intel (approved):
Thanks Adrian, I hope this fixes the flakiness in the CI. I would still like to understand why it sometimes fails - but not always, and if that can happen outside the CI too.

PR #395, opened 2022-02-01
Remove event from CI buttons link

helena-intel:
If you click on a CI button, you now go to a page that doesn't show all CI runs. This PR fixes that: it now shows all CI runs for the main branch.


ryanloney (approved):
Thanks for fixing this, I was wondering why it did not work

PR #401, opened 2022-02-03
Ci 112 speedup

helena-intel:
* Speed up 112 in CI by using a small subset of data in the test (with test_replace)
* Minor code improvement for 302 by removing `fp32_checkpoint_filename` variable and replacing with `fp32_pth_path.name`
* Added 112 and 302 back to CI, since tiny-imagenet is back online. Added tiny-imagenet zipfile to CI cache, and added function that checks that files in cache still exist on the server - that way next time a download fails temporarily, we can still run the nbval tests but the `urls_exists` test warns about the issue so we are aware of it and can take action (this does not at this moment check that cached OMZ files exist, only explicitly downloaded files).

### DIFF
--- a/notebooks/302-pytorch-quantization-aware-training/302-pytorch-quantization-aware-training.ipynb
+++ b/notebooks/302-pytorch-quantization-aware-training/302-pytorch-quantization-aware-training.ipynb
@@ -135,7 +135,6 @@
     "DATA_DIR.mkdir(exist_ok=True)\n",
     "\n",
     "# Paths where PyTorch, ONNX and OpenVINO IR models will be stored\n",
-    "fp32_checkpoint_filename = Path(BASE_MODEL_NAME + \"_fp32\").with_suffix(\".pth\")\n",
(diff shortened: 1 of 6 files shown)


notebooks/302-pytorch-quantization-aware-training/302-pytorch-quantization-aware-training.ipynb
```diff
@@ -182,7 +181,6 @@
     "    url=\"http://cs231n.stanford.edu/tiny-imagenet-200.zip\",\n",
     "    tarname=\"tiny-imagenet-200.zip\",\n",
     "):\n",
-    "    data_dir.mkdir(exist_ok=True)\n",
```
ljaljushkin, line 185: Will it be ok if user runs it for the first time in the clean environment?

helena-intel: Yes, the download_file function creates the directory if it doesn't exist. This is also tested in the CI.

ljaljushkin (approved):

ryanloney (approved):

kshpv (approved):

PR #421, opened 2022-02-16
Fix 202 error when NUM_FRAMES=0

### LINKED ISSUE #402: ZeroDivisionError: float division by zero
When you set `NUM_FRAMES = 0` in 202-vision-superresolution-video it works but there is ZeroDivisionError: float division by zero. Probably `cap.get(cv2.CAP_PROP_FRAME_COUNT)` is also 0.
Screenshot from 2022-02-03 13-40-51

helena-intel:
Fixes #402

Also removed note about model suggestion, updated formatting, and replaces os.makedirs with equivalent pathlib method.


adrianboguszewski (approved):
Thanks! It works now.

PR #432, opened 2022-02-24
Release/1 series update

Debskij:


notebooks/103-paddle-onnx-to-openvino/103-paddle-onnx-to-openvino-classification.ipynb
```diff
@@ -11,22 +11,16 @@
     "\n",
```
paularamo, line 11: I got a timeout error. I know this is because the Hub Server cannot connect with paddlepaddle.org. So, I am wondering if we can include a note to say that this part (PaddlePaddle model downloading) could fault.

paularamo (changes requested):
I was not able to test this notebook, I got a timeout error with PaddlePaddle Hub

notebooks/103-paddle-onnx-to-openvino/103-paddle-onnx-to-openvino-classification.ipynb
paularamo: Error with omz_python module on 110. 

image

paularamo: Question over 112 - I am expecting to have high throughput with INT8. What is the reason we have less throughput with INT8?

image

PR #435, opened 2022-02-24
Release/4 series update

Debskij:


ryanloney:
@Debskij please make these PRs to 2022.1 branch (not main). Also the requirements.txt should not use test.pypi.org. We have the pre-release packages on main PyPI now: https://pypi.org/project/openvino-dev/2022.1.0.dev20220215/ (this is latest)

PR #442, opened 2022-02-25
105 demo and performance comparison

adrianboguszewski:


notebooks/105-language-quantize-bert/105-language-quantize-bert.ipynb
```diff
@@ -522,10 +630,60 @@
...
+    "inputs = data_loader[0][1]\n",
+    "\n",
+    "with torch.no_grad():\n",
+    "    start = time.perf_counter()\n",
+    "    for _ in range(num_samples):\n",
+    "        torch_model(torch.as_tensor(list(inputs.values())).squeeze())\n",
+    "    end = time.perf_counter()\n",
+    "    time_torch = end - start\n",
+    "print(\n",
+    "    f\"PyTorch model on CPU: {time_torch / num_samples:.3f} seconds per image, \"\n",
```
helena-intel, line 659: This should not be "seconds per image". Seconds per sentence?

notebooks/105-language-quantize-bert/105-language-quantize-bert.ipynb
```diff
@@ -522,10 +630,60 @@
...
+    "    f\"FPS: {num_samples / time_torch:.2f}\"\n",
```
helena-intel, line 660: It is called FPS in benchmark_app too, but it is an image specific term.

notebooks/105-language-quantize-bert/105-language-quantize-bert.ipynb
```diff
@@ -522,10 +630,60 @@
     "tags": []
    },
    "source": [
-    "## Compare Performance of the Original and Quantized Models\n",
-    "Finally, we will measure the inference performance of the FP32 and INT8 models. To do this, we use [Benchmark Tool](https://docs.openvinotoolkit.org/latest/openvino_inference_engine_tools_benchmark_tool_README.html) - OpenVINO's inference performance measurement tool.\n",
+    "## Compare Performance of the Original, Converted and Quantized Models\n",
+    "\n",
+    "To see the performance difference we will compare original PyTorch and OpenVINO converted model (FP32)."
```
helena-intel, line 635:
```suggestion
    "To see the performance difference we will compare the original PyTorch model with the OpenVINO quantized model (INT8)."
```
The model loaded is `compressed_model_xml`. It makes sense to me to compare with the INT8 model too. You could also add the IR FP32 or FP16 comparison here, though that's also in benchmark_app right below this function.

notebooks/105-language-quantize-bert/105-language-quantize-bert.ipynb
```diff
@@ -196,11 +198,31 @@
    },
    "outputs": [],
    "source": [
-    "ir_model_xml = onnx_model_path.with_suffix(\".xml\")\n",
-    "ir_model_bin = onnx_model_path.with_suffix(\".bin\")\n",
+    "ir_model_xml_32 = onnx_model_path.with_name(onnx_model_path.stem + \"_32\").with_suffix(\".xml\")\n",
```
helena-intel, line 201: On CPU there is no performance difference between FP16 and FP32. I tried benchmark_app with GPU on my TGL laptop, and there is a nice progression from FP32->FP16->INT8. It is good to show that.
I recommend adding a short explanation here about why we're converting to both FP16 and FP32 and mention in the benchmark_app section that FP16 is only faster on GPU. 
The model loads pretty quickly on GPU so IMO it is useful to add a line for the benchmark_app commands with (untested old API syntax) `device = "GPU" if "GPU" in ie.available_devices else "CPU"` so it will automatically show the best performance on computers with a supported iGPU.

helena-intel (changes requested):
Thank you @adrianboguszewski ! Getting an inference demo for this model was on the top of my improvements wishlist. Very happy to see this.
image

Could you add a short explanation about what the inference does and what kind of data is returned by the dataloader? It is pretty understandable when you see the output, but it is nice to have some context. 

I would have loved to be able to type my own sentences, but that's probably out of scope for this notebook. I do like that it's easy to change the index to get a new pair of sentences.

adrianboguszewski:
That's true there is no difference between FP16 and FP32 models on CPU but I cannot use GPU as I'm using dynamic shapes. In that case, I decided to get rid of FP16. There are original model, FP32 and INT8 now. I think it's enough to show the difference between original, converted and quantized models.

adrianboguszewski:
@helena-intel @ryanloney @KodiaqQ any additional comments or we can merge?

ryanloney (approved):

PR #445, opened 2022-02-28
Update demo requirements file

Debskij:
Pinning requirements to avoid vulnerability. More information in https://github.com/openvinotoolkit/openvino_notebooks/pull/436


helena-intel (approved):
Thanks @Debskij . numpy and scipy are installed by openvino-dev, they seem superfluous in the original file now. You could remove them, as well as opencv-python (also installed by openvino-dev). Keeping them as you did now is fine too.

PR #455, opened 2022-03-04
Temporarily disable Python 3.6 on Windows in nbval

helena-intel:
nbval started failing on Windows/Python 3.6 during requirements installation. Temporarily disable this combination in CI.


helena-intel:
See here for an example failed run: https://github.com/openvinotoolkit/openvino_notebooks/runs/5431929803?check_suite_focus=true
 
 error: the configured Python interpreter version (3.6) is lower than PyO3's minimum supported version (3.7)

PR #462, opened 2022-03-11
Update System Requirements OpenVINO 2022.1 (Readme File)

paularamo:
- Add an important note: Users should be aware about the branch update.

- Python 3.10 is not supported.

- Python 3.9 added into the System Requirements table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ English | [简体中文](README_cn.md)
 
 A collection of ready-to-run Jupyter notebooks for learning and experimenting with the OpenVINO™ Toolkit. The notebooks provide an introduction to OpenVINO basics and teach developers how to leverage our API for optimized deep learning inference.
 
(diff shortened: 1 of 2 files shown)


adrianboguszewski:
"For Python 3.9, C++ is also required" should be removed? Is it only for Windows?

ryanloney:
Installing editdistance (dependency of dependency) used to required C++ with Python 3.8 and higher. This may not be required anymore. I don't know. 

@helena-intel do you happen to know? Otherwise we should test to confirm @paularamo

helena-intel:
> Installing editdistance (dependency of dependency) used to required C++ with Python 3.8 and higher. This may not be required anymore. I don't know.
> 
> @helena-intel do you happen to know? Otherwise we should test to confirm @paularamo

editdistance is no longer required (CVS-69819) so unless some other package now requires C++, this should no longer be necessary.

PR #469, opened 2022-03-14
Update notebook utils

Debskij:
Update type hints and fix default values in utils


notebooks/utils/notebook_utils.py
```diff
@@ -623,9 +623,9 @@ def show_live_inference(ie, image_paths: List, model: model.Model, device: str):
 
 def benchmark_model(model_path: os.PathLike,
```
helena-intel, line 624:
```suggestion
def benchmark_model(model_path: PathLike,
```
Also changed in ipynb file, fix inconsistency in existing code.

helena-intel (approved):
Thanks @Debskij ! 
I now realize that functions that accept either a Path or a str should technically have a type hint of Union[str, PathLike] (https://peps.python.org/pep-0519/#provide-specific-type-hinting-support) but for now, IMO we can leave it as is.

PR #476, opened 2022-03-16
Update README with 2022.1 upgrade instructions

ryanloney:


adrianboguszewski:
@ryanloney cannot merge it without admin permissions

PR #478, opened 2022-03-17
Added an example of resnet-50-pytorch model

tiny-dentist:
This PR adds support for openvino fp16 inference for pytorch-resnet-50 model. Successfuly converted it to fp16 format and tried it on Device CPU and GPU with model accuracy result verified.

The flow of the model is
block_digram


rahulsingh701:
@adrianboguszewski could you please run the workflow test again?

AlexKoff88:
Thanks for the contribution!

But I am not sure about the value of this notebook because we already have a very similar [one](https://github.com/openvinotoolkit/openvino_notebooks/blob/main/notebooks/001-hello-world/001-hello-world.ipynb). It looks like duplication to me. @helena-intel, @ryanloney, please take a look as well.

rahulsingh701:
Hello @AlexKoff88 , Thanks for your reply. I am sorry I did not know that my notebook and the one you had tagged via the link are similar. I did not mean to do this. Should I create a new PR with a different model not present in the openvino repository?

AlexKoff88:
The notebook, which could make sense to me, is, for example, inference and optimization (quantization) of any model from timm. It could be ResNet-18. It would be helpful to show end-to-end workflow from model export to ONNX, then to IR, quantization, performance benchmarking, results on real data.

adrianboguszewski:
@rahulsingh701 @AlexKoff88 is right. This notebook is a duplication. We accept it as a pre-requisite task. You don't have to create a new one. But of course, if you want you can.

PR #503, opened 2022-03-25
Adding 221-deblurgan-v2

jainsid2305:
This is my submission for the Prerequisite task for contributing in Openvino for GSoC.
In This Notebook, I have implemented the deblurgan-v2 deblurring model with Openvino.

Screenshot (311)


raymondlo84:
Should avoid the binary files if they are going to be generated.

jainsid2305:
> Should avoid the binary files if they are going to be generated.

Thanks. I have removed the binary files. Can you have a look now if there are any further changes?

Debskij (changes requested):
Please remove model files from repo. It should be downloaded by notebook itself.

PR #515, opened 2022-03-29
Fix catching HTTP error in download_file

helena-intel:
In download_file URLError was caught first, with an if statement for timeout errors. HTTPError is a subclass of URLError and was never caught. This PR fixes that, by catching the more specific HTTPError first, and will also raise an exception now on non-timeout related other URLErrors.

Previously, if download_file tried to download a URL that does not exist, the function would fail without a clear indication of what went wrong. Now the expected error message is shown:
06836cea36df6644e9903493f12fb065

There are also some formatting changes (by black). The only content changes are line 139-151 in the notebook.


ryanloney:
@Debskij @paularamo please review and merge when possible

paularamo (approved):

Debskij (approved):

PR #527, opened 2022-04-06
[GSOC]225-Resnet18 QAT with weights obtained from Pytorch Image Model

tiny-dentist:
Added an example of pretrained Resnet18 model with weights obtained from Pytorch-Image-Model. Fine tuned for training on custom dataset. Applied Quantization Aware Training to it and exported both FP32 and INT8 model to ONNX for conversion to OpenVINO IR format. Benchmarked both the models on 1000 images and calculated time required and FPS obtained.

FP32 benchmark:

FP32_benchmark

INT8 Benchmark:

INT8_benchmark


tiny-dentist:
@AlexKoff88 Here is the notebook I was talking about using TIMM to obtain a pretrained Resnet18 model which I later trained on a custom dataset and used QAT to then convert it into INT8 and into OpenVINO IR format

adrianboguszewski:
Hi @tiny-dentist. Thank you for your contribution. As this is part of one of our GSoC projects which has not been implemented this year, we decided to close it.

PR #529, opened 2022-04-06
[GSOC] 226-yolo-v4-tf object detection notebook.

thavens:
A notebook that implements Yolo-v4-tiny-tf and yolo-v4-tf. Compared to the 401 object detection notebook, changes had to be made to the output processing to find bounding boxes and to resize the image while preserving the aspect ratio for improved performance.
[yolov4 model](https://docs.openvino.ai/2021.2/omz_models_public_yolo_v4_tf_yolo_v4_tf.html)

What's left:
Some documentation / explanation.

Something I found that needs to be confirmed is that Cx, Cy (cell index) and w, h (bounding box width/height), from the documentation, needs to have order changed to Cy, Cx and h, w respectively.
Converted model input documentation should also become B, H, W, C, instead of B, C, H, W which gives an error. Whether or not the input is BGR or RGB isn't too clear yet considering the model goes by original input for dimensions.

Edit:
[New Documentation](https://docs.openvino.ai/latest/omz_models_model_yolo_v4_tf.html) is consistent with the current inputs I have. So BHWC is correct (using BGR).


raymondlo84:
Do you have a screenshot / video of it running?

thavens:
[yolo v4 tiny bgr](https://youtu.be/2rXpEB0APM4)
[yolo v4 tiny rgb](https://youtu.be/ucONs1E96pY)
Even looking frame by frame, they have a difference but it's not obvious enough to point out visually.

raymondlo84:
You can try things that are more sensitive to color channels... like apple, orange etc... ;)

    'person', 'bicycle', 'car', 'motorcycle', 'airplane', 'bus', 'train', 'truck', 
    'boat', 'trafficlight', 'firehydrant', 'stopsign', 'parkingmeter', 'bench', 'bird', 
    'cat', 'dog', 'horse', 'sheep', 'cow', 'elephant', 'bear', 'zebra', 'giraffe', 
    'backpack', 'umbrella', 'handbag', 'tie', 'suitcase', 'frisbee', 'skis', 'snowboard', 
    'sportsball', 'kite', 'baseballbat', 'baseballglove', 'skateboard', 'surfboard', 
    'tennisracket', 'bottle', 'wineglass', 'cup', 'fork', 'knife', 'spoon', 'bowl', 
    'banana', 'apple', 'sandwich', 'orange', 'broccoli', 'carrot', 'hotdog', 'pizza', 
    'donut', 'cake', 'chair', 'couch', 'pottedplant', 'bed', 'diningtable', 'toilet', 
    'tv', 'laptop', 'mouse', 'remote', 'keyboard', 'cellphone', 'microwave', 'oven', 
    'toaster', 'sink', 'refrigerator', 'book', 'clock', 'vase', 'scissors', 
    'teddybear', 'hairdrier', 'toothbrush'
]

thavens:
Super smart idea! I just tried it and can confirm the model is converted to BGR which is in line with docs. Also, I'm more confident in saying that the output documentation is wrong. I can't find anywhere I'd be wrong here. Maybe during review, you might want to look a bit closer at the result processing because accepting would imply the documentation is wrong.

thavens:
PR should be good to go now!

thavens:
Build_nbval seems to fail on clear output. Is there any way to fix that or should I just change the default functionality to open a new cv2 window instead? The default video display code is from notebook 401 though, and it runs fine on my ubuntu 20.04.

raymondlo84:
> Super smart idea! I just tried it and can confirm the model is converted to BGR which is in line with docs. Also, I'm more confident in saying that the output documentation is wrong. I can't find anywhere I'd be wrong here. Maybe during review, you might want to look a bit closer at the result processing because accepting would imply the documentation is wrong.

:D When you got stuck enough -- you will start grabbing whatever it takes to debug. :)

raymondlo84:
> Build_nbval seems to fail on clear output or no output. Is there any way to fix that or should I just change the default functionality to open a new cv2 window instead? The default video display code is from notebook 401 though, and it runs fine on my ubuntu 20.04. image

It should not use the pop up for the ci? I would see what @adrianboguszewski would say about this

thavens:
[check](https://github.com/pytube/pytube/issues/1299)
[fix](https://stackoverflow.com/questions/68945080/pytube-exceptions-regexmatcherror-get-throttling-function-name-could-not-find/71903013#71903013)
There seems to be an issue with PyTube and Cypher for regex search. Youtube changed something on their side and I'd have to modify the library or wait for the fix to be put on pip or somehow remove the dependency on pytube.

thavens:
Ok, so I tried using the GUI-based popup for OpenCV to output images. This fixed the issue on systems with GUI but now it fails on the systems which I assume have no GUI. I'm running this on ubuntu 22.04 python 3.8 w/ GUI so I know it works.
I think notebook 401 passed the tests without a workaround because it used webcam input and didn't trigger the automatic error when there is no output within 5 seconds.

Docker build failed because it doesn't seem to have the updated pytube library. Maybe someone can update the docker image to pytube 12.1.0.

adrianboguszewski:
Hi @thavens. Thank you for your contribution. This notebook is very similar to 401. Moreover, we're going to add Yolo v5, so we decided not to merge this.

PR #530, opened 2022-04-07
Update piplock

Debskij:


helena-intel (approved):
Thanks @Debskij! Is this a revert to a previous Pipfile.lock? I noticed in the diff that many packages have lower versions now than in the current version. 

Merging this now, since with existing Pipfile.lock CI fails for Docker, and CI passes with this PR.

PR #534, opened 2022-04-09
update the 214 benchmarkapp command

OpenVINO-dev-contest:


ryanloney:
@OpenVINO-dev-contest can you explain the gap between your results and the benchmark app? For latency mode I'm seeing 145fps in your loop and 323fps in benchmark_app. 
Screenshot from 2022-04-09 21-04-41
vs.
Screenshot from 2022-04-09 21-07-08

For throughput mode, I'm seeing 
Screenshot from 2022-04-09 21-08-01

vs. 
Screenshot from 2022-04-09 21-08-31

OpenVINO-dev-contest:
> @OpenVINO-dev-contest can you explain the gap between your results and the benchmark app? For latency mode I'm seeing 145fps in your loop and 323fps in benchmark_app. Screenshot from 2022-04-09 21-04-41 vs. Screenshot from 2022-04-09 21-07-08
> 
> For throughput mode, I'm seeing Screenshot from 2022-04-09 21-08-01
> 
> vs. Screenshot from 2022-04-09 21-08-31

Good question. I think it because Benchmark will convert the only one input image/dummy data to ov.tensor, and save it in a shared memory for each inference. But in real situation, the input image will be changed for every loop, and the cost of memory copy and data converting should be measured.

ryanloney (approved):

PR #539, opened 2022-04-19
	223-text-prediction

dwipddalal:
# Interactive Text Prediction with OpenVINO
This is a demo for text prediction using [gpt-2](https://github.com/openvinotoolkit/open_model_zoo/tree/master/models/public/gpt-2) model
The complete pipeline of this demo's notebook is shown below.

image2

</br>
This is an interactive demonstration in which the user can type text into the input bar and generate predicted text. This procedure can be repeated as many times as the user desires.
</br>
</br>

image3

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,8 @@ More amazing notebooks here!
 | [111-detection-quantization](notebooks/111-detection-quantization/)<br>[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/openvinotoolkit/openvino_notebooks/HEAD?filepath=notebooks%2F111-detection-quantization%2F111-detection-quantization.ipynb) | Quantize an object detection model | 
 | [112-pytorch-post-training-quantization-nncf](notebooks/112-pytorch-post-training-quantization-nncf/) | Use Neural Network Compression Framework (NNCF) to quantize PyTorch model in post-training mode (without model fine-tuning)| 
 | [113-image-classification-quantization](notebooks/113-image-classification-quantization/) | Quantize mobilenet image classification | 
-| [114-quantization-simplified-mode](notebooks/114-quantization-simplified-mode/) | Quantize Image Classification Models with POT in Simplified Mode| |
+| [114-quantization-simplified-mode](notebooks/114-quantization-simplified-mode/) | Quantize Image Classification Models with POT in Simplified Mode|
+| [115-async-api](notebooks/115-async-api/) | Use Asynchronous Execution to Improve Data Pipelining| |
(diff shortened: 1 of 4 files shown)


dwipddalal:
Is this PR ready to be merged, or has it already been merged?

adrianboguszewski:
Hi @dwipddalal. Thank you for this great notebook. We're sorry for the waiting time. We're a small team and were very busy. Absolutely we would like to merge it. It should be possible in July/August. Please wait for the reviews.

adrianboguszewski (changes requested):
This is really great and useful notebook. However, we need to change some things to be able to merge it. Please:
- change the name to 223-text-prediction (new number and lowercase)
- change the extension of README from txt to md
- add this notebook to the main README and notebooks/README files
- remove HTML tags from README.txt and replace them with markdown
- remove DEVICE var as the only option in case of dynamic shapes is CPU
- add this note above dynamic shapes assignment (Note: Since the text recognition model is with dynamic input shape and current release of OpenVINO 2022.1 does not support dynamic shape on iGPU, you cannot directly switch device to iGPU for inference in this case. Otherwise, you may need to resize the input images to this model into a fixed size and then try running the inference on iGPU.)
- check if the model is converted, not to convert many times (see e.g. 401)
- displaying download commands is in the wrong cell
- print the name and shape of input and output nodes instead of objects
- use a longer sequence than 64 e.g. 128
- remove interactive mode as it causes issues with our CI (remove the last while loop and input, assign the beginning to a variable instead)

dwipddalal:
Sure @adrianboguszewski, I shall immediately start working on these improvements. Thank you.

dwipddalal:
Do I have to add this notebook in the following section?
image

adrianboguszewski:
Not this section, but add it to the table below:
image

dwipddalal:
Okay sure.

adrianboguszewski:
Hi @dwipddalal, could you explain what happened here? This situation (merging 0 commits to our main branch without CIs passing and by an unauthorized person) is impossible taking into account our settings.

dwipddalal:
I pushed back my commits and updated my branch to the date so I could make a new pull request with all the changes you suggested. 
I guess there is some fault in the system because of which it is showing "0 commits to our main branch."

dwipddalal:
I made a new pull request #611

adrianboguszewski:
Thanks. Indeed, it looks like a Github glitch/issue.

dwipddalal:
Yes. Please do have a look at new PR and let me know if I have to make any more changes or is it ready?

PR #574, opened 2022-06-17
Add NNCF kidney segmentation quantization notebook

helena-intel:
This notebook requires a C++ compiler (like 302). The function that tries to find the path to cl.exe in 302 does not work well in all cases on Windows 11. Users should start Jupyter from a Developer Command Prompt - then cl.exe should be found. If importing torch fails, it is usually because of C++ issues. I added a warning if that happens.

There is an issue with the NNCF model: it is not quantized the same way as the model in the POT notebook. This is why it is required to add the `ignored_scopes` to the config to allow the model to run on GPU. Inference is also a bit slower on this model than on the POT model. 

I did not add this notebook to the README yet.


notebooks/110-ct-segmentation-quantize/110-ct-segmentation-quantize-nncf.ipynb
```diff
@@ -0,0 +1,795 @@
...
+    "except subprocess.CalledProcessError:\n",
+    "    message = \"WARNING: Please install a C++ compiler to use this notebook.\"\n",
+    "    if sys.platform == \"win32\":\n",
+    "        message += (\n",
+    "            \" If you installed Visual Studio Build Tools, please open a Developer \"\n",
+    "            \"Command Prompt and start Jupyter from there.\"\n",
+    "        )\n",
+    "    NotebookAlert(message=message, alert_class=\"warning\")\n",
+    "    raise"
+   ]
```
paularamo, line 94: Windows 11. Microsoft Visual Studio 2022 Community. Running in Developer Command prompt. cl.exe path added into environmental variables. Python version 3.9.13. I got this error.

image

helena-intel: It seems tricky to get this right on Windows. I think the developer command prompt works well if you use the x64 prompt and the version of Python that is installed with Visual Studio, but that's not something that most Python developers will be familiar with. I added the pathfinder cell back to the top, and removed the recommendation to use a developer prompt. 

I added a link to VS install instructions to the notebook (and also mentioned what to install for macOS and Linux). OpenVINO system requirements specify Visual Studio 2019, and that's what we test, so I added a link to that. It may work well with VS 2022 though - but I cannot test that at the moment.

The error message in the screenshot shows that the default installation root for 2022 is different. I would try this: start Jupyter from a regular, non developer prompt, and change the VS_INSTALL_DIR line in the pathfinder cell to:

`VS_INSTALL_DIR = r"C:/Program Files/Microsoft Visual Studio"`

paularamo (approved):
@ryanloney could you please review this notebook and merge. We need this notebook for tomorrow. Thanks. No issues with Win 11, working on Ubuntu and Win 10.

ryanloney (approved):

helena-intel:
Currently notebook works with VS 2019 (OpenVINO supported VS version) on Windows 10 and 11. 

PR with VS 2022 support: https://github.com/openvinotoolkit/openvino_notebooks/pull/577

PR #587, opened 2022-07-04
Update new demo 219-pytorch-object-detection-accuracy-check-and-quantization

FionaZZ92:
[219-pytorch-object-detection-accuracy-check-and-quantization](https://github.com/FionaZZ92/openvino_notebooks/blob/main/notebooks/219-pytorch-object-detection-accuracy-check-and-quantization/README.md)


FionaZZ92:
My demo code is based on 2022.1 OV POT API, but docker env still use OV2021.4. 
Thus, I ignore the docker env check with my demo 219-pytorch-object-detection-accuracy-check-and-quantization.

On the other hand, I add seaborn in requirements.txt, which not added in docker env pipFile. I think if the demo not check in docker env, no need require same pkg with project py env.

Debskij:
Hi @FionaZZ92! As you can see in https://github.com/openvinotoolkit/openvino_notebooks/blob/main/.docker/Pipfile#L25 Docker tests are running on the latest 2022.1 OpenVINO, therefore there is no reason to ignore it for your notebook. 
Please rebase your branch to latest main, update pipfile and pipfile.lock by adding seaborn, and add remove your notebook from the ignored list in docker validation.

PR #591, opened 2022-07-14
Modify runner os names

Debskij:


paularamo (approved):

ryanloney:
@Debskij what is the reason for changing the runner names?

ryanloney (approved):

PR #602, opened 2022-08-03
Replace deprecated MacOS version with latest 

Debskij:


adrianboguszewski:
Are we sure we want to remove macos 10.15? It's not EOL according to this site: https://endoflife.date/macos. Moreover, I think we should add macos 12.

adrianboguszewski (approved):

paularamo (approved):

PR #604, opened 2022-08-05
Add teardown for nbval testing

Debskij:
Create a script that takes care to remove unnecessary content from already tested notebooks. This fixes no space left on the device.

### DIFF
--- a/.ci/validate_notebooks.sh
+++ b/.ci/validate_notebooks.sh
@@ -1,18 +1,23 @@
 ignorelist=""
 failed=0
 earlystop=false
+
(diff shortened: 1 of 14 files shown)


adrianboguszewski:
This solution works for too long (180min vs 60min) and it passes even though some notebooks are failing. More work needs to be done.

adrianboguszewski (changes requested):
This solution works for too long (180min vs 60min) and it passes even though some notebooks are failing. More work needs to be done.

adrianboguszewski:
Work continued in #607

PR #609, opened 2022-08-10
Pytube Upgrade

### LINKED ISSUE #403: YouTube support broken in 202 video superresolution notebook
PyTube is broken at the moment. It is only used in 202 and already in a try/except loop so the notebook still works with the default video, but it currently does not work with YouTube videos.  https://github.com/yt-dlp/yt-dlp looks like a good alternative.

AnishaUdayakumar:
Updated requirements.txt with pytube >=12.1.0 to support superresolution notebook. Fixes #403


adrianboguszewski:
The work is continued in #619

PR #619, opened 2022-08-25
Piplock updates + ubuntu 18 and python 3.6 depreaction

### LINKED ISSUE #403: YouTube support broken in 202 video superresolution notebook

adrianboguszewski:
It removes Ubuntu 18.04 and python 3.6 from our CI. It fixes all problems with Pipfile. Fixes #403 as well


ryanloney (approved):

Debskij (approved):
Good job!

PR #620, opened 2022-08-25
Requirements lock

adrianboguszewski:
This PR pins all requirements to a specific (working) version. It should fix "morning surprises" that some package was automatically updated overnight and our notebooks stopped working.

This improves the stability of the project.


.docker/Pipfile
```diff
@@ -9,51 +9,62 @@ verify_ssl = true
 name = "pytorch-wheels"
 
 [packages]
-boto3 = "*"
-fastseg = "*"
-gdown = "*"
+addict = "==2.4.0"
```
zhuo-yoyowz, line 12: what's use of the package "addict"? why we need add it here?

zhuo-yoyowz (approved):

.docker/Pipfile
adrianboguszewski: It is used in some notebooks e.g. 110 or 113, but this is a "weak" dependency. I assume it could be easily replaced in the future by ordinary Python dict.

raymondlo84:
Do we have a 'relaxed' version? For example, we had to hard lock the Tensorflow before because it was breaking the code, but we try to relax the other so we can get the security updates?

raymondlo84:
Also, how are we going to ever update? We are going to change each package's version manually?

adrianboguszewski:
@raymondlo84 the package versions come from the relaxed version of requirements. I mean I removed as many requirements as possible, allowed pip to install the best versions matching all remaining restrictions and then assigned the versions in the requirements file.

The idea for the update is the same. Relax everything what possible, allow pip to install the best versions, and assign exact versions to packages once again.

This is my proposition on how to make the repository more stable. It requires further discussion with the team.

PR #622, opened 2022-08-26
Removed ignoring notebooks for mac

adrianboguszewski:
This is updating 107 and 115 to mitigate timeouts on mac


Debskij (approved):

adrianboguszewski:
This PR no longer makes sense. There is a bug with async mode on mac and nothing we do can help.

PR #624, opened 2022-08-30
Fixed links for badges

adrianboguszewski:
Fixed links for badges


ryanloney (approved):
Thanks for fixing this

paularamo (approved):

PR #626, opened 2022-09-06
Fixes wrong data type

### LINKED ISSUE #625: 220 notebook mentions FP32 IR in the accuracy comparison but should be FP16
**Describe the bug**
The description everywhere in the notebook 220-yolov5-accuracy-check-and-quantization says about FP32 vs INT8 but it should be FP16 vs INT8 because MO creates FP16 IR.

 **Expected behavior**
Change the MO conversion to FP32: --data_type FP32
 or
Use FP16 instead of FP32 everywhere in the description (incl README)

adrianboguszewski:
Fixes #625


AnishaUdayakumar (approved):
Notebook runs successfully after the data type changes

paularamo (approved):

PR #644, opened 2022-10-07
Fixed the order of performance benchmarks

adrianboguszewski:
This changes the order of performance benchmarks to more reasonable


AnishaUdayakumar (approved):
Tested the order change and it works as expected.

zhuo-yoyowz (approved):

PR #658, opened 2022-11-11
Remove API 1.0 usage from remaining notebooks

eaidova:


ryanloney (approved):

paularamo (approved):

adrianboguszewski (commented):
async_pipeline.py, custom_segmentation.py and model.py are not widely used (only in 110 and 210). In my opinion, they should be moved to these directories or even the content should be merged into those notebooks to improve understandability. I suggest doing it as a part of this PR.

eaidova:
> async_pipeline.py, custom_segmentation.py and model.py are not widely used (only in 110 and 210). In my opinion, they should be moved to these directories or even the content should be merged into those notebooks to improve understandability. I suggest doing it as a part of this PR.

I'm not sure that I want spend time for it, because one of the next my step is remove this api demonstration at all. Regarding model classes, now they have nothing common with Model API interface except name, it may mislead users. Regarding AsyncPipeline, OpenVINO 2.0 brings easy to use solution for async pipelines construction - AsyncInferQueue, I think it is better to use it, instead of self-written async pipeline in style of previous api.

adrianboguszewski (approved):
Ok. So I'm fine.

PR #693, opened 2022-12-02
Add script for checking link

sbalandi:


.ci/check_links.py
```diff
@@ -0,0 +1,73 @@
...
+                complain(f'{md_path}: invalid URL reference {url!r}')
+                continue
+
+            if not components.path: # self-link
+                continue
+
+            if not components.scheme and not components.netloc:
+                # check if it is relative path on file from repo
+                file_name = os.path.join(os.path.dirname(md_path), components.path)
+                if not os.path.exists(file_name):
```
eaidova, line 60:
```suggestion
                file_name = md_path.parent / components.path
                if not file_name.exists():
```

.ci/check_links.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python3
+
+"""
+This script is like check-basics.py, but specific to the documentation.
```
eaidova, line 4: probably, this comment not related to this script, because we have not check_basic.py :)

.ci/check_links.py
eaidova: after that os package should be removed from imports

.ci/check_links.py
```diff
@@ -0,0 +1,66 @@
...
+
+            if not components.scheme and not components.netloc:
+                # check if it is relative path on file from repo
+                file_name = md_path.parent / components.path
+                if not file_name.exists():
+                    complain(f'{md_path}: invalid URL reference {url!r}')
+                continue
+
+            try:
+                get = requests.get(url, proxies={"http": "http://proxy-chain.intel.com:911", 'https': 'http://proxy-chain.intel.com:912'})
```
sbalandi, line 58: @eaidova I add proxy here, Is it right way or proxy will be added by ci steps ?

eaidova: ci scripts running on hosts outside organization, so it should not have proxy settings, I think if they are required, they should be setup manually.

sbalandi: Thank you, removed

adrianboguszewski (approved):
This is a really great idea!

PR #673, opened 2022-11-20
102 replace fastseg model

eaidova:

PR #697, opened 2022-12-06
update torch one more time

eaidova:


eaidova:
ported to https://github.com/openvinotoolkit/openvino_notebooks/pull/673

PR #700, opened 2022-12-06
use upperbound for transformers for prevent gpt2 conversion issue

eaidova:
workaround until https://github.com/openvinotoolkit/open_model_zoo/pull/3621 is not a part of openvino-dev


adrianboguszewski (commented):
What about pipfile? Should it be there also?

eaidova:
> What about pipfile? Should it be there also?

we have 4.24.0 in pipfile.lock, so it should not have issues, but of course I will update pipefile too.

adrianboguszewski (approved):

PR #710, opened 2022-12-12
prevent out of memory issue for windows in precommit

eaidova:


.ci/validate_notebooks.py
```diff
@@ -59,13 +59,25 @@ def prepare_test_plan(test_list, ignore_list):
     return statuses
 
 
+def clean_test_artefacts(before_test_files, after_test_files):
+    for file_path in after_test_files:
+        if file_path in before_test_files or not file_path.exists():
+            continue
+        if not file_path.is_file():
```
adrianboguszewski, line 66: shouldn't be "if file"?

.ci/validate_notebooks.py
```diff
@@ -59,13 +59,25 @@ def prepare_test_plan(test_list, ignore_list):
...
+            file_path.unlink()
+        else:
+            file_path.rmdir()
```
adrianboguszewski, line 69: Will it work for directories with files inside?

eaidova: replaced with another command

adrianboguszewski (approved):

PR #722, opened 2022-12-20
Update with the PaddleOCR V3 model

zhuo-yoyowz:
Update this notebook with the new PaddleOCR v3 model, which provides better text detection and recognition results with a bit larger model size and nearly the same FPS compared to the current PaddleOCR v2 model.


eaidova:
@zhuo-yoyowz I think from educational perspective, it can be useful if you provide info which model used with link on introduction page:
https://github.com/PaddlePaddle/PaddleOCR/blob/release/2.6/doc/doc_en/ppocr_introduction_en.md#pp-ocrv3

eaidova:
Also to prevent timeout issue for test on mac, could you please add files for downloading to pipeline?
https://github.com/openvinotoolkit/openvino_notebooks/blob/main/.github/workflows/nbval_precommit.yml#L156

eaidova (approved):

adrianboguszewski (approved):

PR #725, opened 2022-12-20
move nncf installation to notebooks

eaidova:


notebooks/110-ct-segmentation-quantize/110-ct-segmentation-quantize-nncf.ipynb
```diff
@@ -42,6 +42,18 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "try:\n",
+    "    import nncf  # noqa: F401\n",
+    "except ImportError:\n",
+    "    !pip install git+https://github.com/openvinotoolkit/nncf.git#egg=nncf"
```
adrianboguszewski, line 54: Don't we need to import nncf once again after installing it? And the same comments for similar changes in other notebooks.

eaidova: >Don't we need to import nncf once again after installing it?

I believe not, it is just check that package installed, the code which uses nncf import it in own cells.
I do not want mix up installation and notebooks imports because I believe it is temporal requirements to have it inside notebooks, we can return it back to requirements.txt when nncf 2.4 release happens

adrianboguszewski: Understand

adrianboguszewski (approved):

PR #733, opened 2022-12-22
try workaround for memory issue on windows

eaidova:


adrianboguszewski (commented):
What about the solution proposed [here](https://github.com/actions/runner-images/issues/1341#issuecomment-1332779364) (working-directory option)?

eaidova:
> What about the solution proposed [here](https://github.com/actions/runner-images/issues/1341#issuecomment-1332779364) (working-directory option)?

this solution will requires to introduce it for all stages, use custom checkout stage and separate windows pipeline with duplication (or have if-statement  for windows for each stage), which can be hard to maintain in future... My solution is localized the most memory consuming part execution without requirement to isolate pipeline

eaidova:
Looks like it works!
But need prevent several additional instabilities

PR #739, opened 2022-12-27
Fixed parsing of benchmark app's output

ljaljushkin:
Changed the rule for parsing output from benchmark app.
Previous rule removed all lines including essential information about performance:
```
Benchmark FP32 model (OpenVINO IR)
Benchmark INT8 model (OpenVINO IR)
```
The ouptut is the following after the fix:
```
Benchmark FP32 model (OpenVINO IR)
[ INFO ] Throughput:   1319.79 FPS
Benchmark INT8 model (OpenVINO IR)
[ INFO ] Throughput:   4780.27 FPS
```


eaidova (approved):

eaidova:
@ljaljushkin  thank you!

Issue #684, opened 2022-11-28
Apple M1/M2 Bug lists


raymondlo84:
Keep tracking a set of bugs we face when we run OpenVINO 2022.3 (arm plugin) with OpenVINO Notebooks here. Since the arm_plugin is not officially supported and thus all fixes here are worked on as case by case only. 

To learn more how to install OpenVINO on Apple Silicons, please read the blog below. 
https://raymondlo84.medium.com/how-to-install-openvino-notebooks-on-apple-m1-m2-silicons-a4c704747c12

raymondlo84:
Known issues:
11/28/2022
1. Human pose estimations 402 is not working 
2. NNCF dependencies in requirements.txt prevented the Notebooks to be installed properly and thus disabled.
3. Performance seems to be even lower than (Rosetta counter part) -- thus this will need to be confirmed.
4. No GPUs support is enabled at the moment.

eaidova:
@raymondlo84 does it still actual?
Probably N2 should be fixed now and nncf installation disabled for notebooks where it does not required, but if you still have any problem with that, let's discuss (possibly with nncf team involving)

what about problem 1.

3-4, I believe are not bugs, it is just specific of experimental arm platform support on openvino side (only cpu plugin enabled, backend libraries difference)

raymondlo84:
I think the main stopper now is the tensorflow. with M1, we have to install tensorflow-macos prior running the pip install -r requirements.txt or we will get this error.


ERROR: Could not find a version that satisfies the requirement tensorflow<=2.9.3,>=2.5; extra == "tensorflow2" (from openvino-dev[onnx,tensorflow2]) (from versions: none)
ERROR: No matching distribution found for tensorflow<=2.9.3,>=2.5; extra == "tensorflow2"

raymondlo84:
ERROR: Could not find a version that satisfies the requirement tensorflow-macos==2.9.3 (from versions: 2.5.0, 2.6.0, 2.7.0, 2.8.0, 2.9.0, 2.9.1, 2.9.2, 2.10.0, 2.11.0)
ERROR: No matching distribution found for tensorflow-macos==2.9.3 

...

raymondlo84:
@adrianboguszewski  -- you have to M1 now and please also feel free to repeat the work.

Also, we didn't specify if we shall use brew or just native python from xcode... So that's another tricky part as it will change our CI too.

raymondlo84:
[...]

PR #766, opened 2023-01-09
New order of requirements, fixing requirements for macOS

adrianboguszewski:
The requirements file was reordered to reduce the clutter. There are also fixes for M1 macOS (#684)

### DIFF
--- a/.docker/Pipfile
+++ b/.docker/Pipfile
@@ -49,7 +49,7 @@ setuptools = ">56.0.0"
 shapely = ">=1.7.1"
 supervisor = "==4.1.0"
 tensorflow = ">=2.5,<=2.9.3"
-tensorflow-datasets = "*"
+tensorflow-datasets = "==4.2.0"
 torch = {index = "pytorch-wheels", version = "==1.13.1+cpu"}
 torchmetrics = ">=0.11.0"
 torchvision = {index = "pytorch-wheels", version = "==0.14.1+cpu"}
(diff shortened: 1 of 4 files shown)


raymondlo84:
tensorflow-macos>=2.5,<=2.9.3; sys_platform == 'darwin'  (this may fail on the x86 mac).

eaidova (approved):

.docker/Pipfile
```diff
@@ -23,35 +23,38 @@ librosa = ">=0.8.1"
...
 rsa = ">=4.7"
 scikit-image = ">=0.19.2"
+scipy = "*"
 seaborn = ">=0.11.0"
 setuptools = ">56.0.0"
 shapely = ">=1.7.1"
 supervisor = "==4.1.0"
-tensorflow_datasets = "*"
+tensorflow = ">=2.5,<=2.9.3"
+tensorflow-datasets = "*"
```
eaidova, line 52: please pin tensorflow-datasets to 4.2.0 for docker, there is some issues with downloading dataset in 305 notebook for latest versionh

eaidova: or probably we should update tensorflow-datasets version in requirements.txt and use `imagenette/160px-v2`, imagenette/160px was removed in newer versions, that is why it can not be downloaded
https://www.tensorflow.org/datasets/catalog/imagenette

adrianboguszewski: I decided to go with 4.2.0 now to mitigate the risk of incompatible dependencies. But I agree it should be updated in the future.

PR #790, opened 2023-01-20
109 performance tricks

adrianboguszewski:
This notebook aims to show some tricks in OV to increase performance


notebooks/109-performance-tricks/109-latency-tricks.ipynb
```diff
@@ -0,0 +1,460 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "num_cores = os.cpu_count()\n",
+    "\n",
+    "ov_cpu_config_model = core.compile_model(ov_model, device_name=\"CPU\", config={\"INFERENCE_NUM_THREADS\": num_cores})\n",
```
OpenVINO-dev-contest, line 286: Hi @adrianboguszewski since sometimes our customers may wish to free the some CPU cores for other non-inference task, this notebook is a real good example to show how to set the configuration for a device. Could you help to show more example about how to set more parameters, for example, the number of stream ? Thanks

adrianboguszewski: I tried to use a different number of streams, but it didn't give me a better performance. Have you tried it? Did you get lower latency?

OpenVINO-dev-contest: Yes, It will not improve the performance. I just want to have a example to show how the set these kind of parameter, cuz many developers are asking about it, and we have no page to explain it except the API document.

adrianboguszewski: I see. I can mention other parameters also, but this notebook should be focused on performance tricks as users are not aware of many of them.

OpenVINO-dev-contest: Understand, changing the number of streams can help to balance the latency and throughout manually. Sometimes we cant just use performance hint to get a best latency or throughout.

notebooks/109-performance-tricks/109-latency-tricks.ipynb
```diff
@@ -0,0 +1,407 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# todo make it 1000\n",
+    "INFER_NUMBER = 100\n",
+    "\n",
+    "def benchmark_model(model, input, model_name, device=\"CPU\"):\n",
```
OpenVINO-dev-contest, line 127: My understanding is this method is used to calculate the latency in Sync mode, but not for latency when runtime is in Async mode. Because in Async mode, there are many parallel tasks, and we only record the total time, not each's task's time.

notebooks/109-performance-tricks/109-latency-tricks.ipynb
```diff
@@ -0,0 +1,407 @@
...
+   "outputs": [],
+   "source": [
+    "from openvino.runtime import AsyncInferQueue\n",
+    "\n",
+    "def callback(infer_request, info):\n",
+    "    # put your post-processing here\n",
+    "    pass\n",
+    "\n",
+    "infer_queue = AsyncInferQueue(ov_auto_model)\n",
+    "infer_queue.set_callback(callback)  # set callback to post-process results\n",
```
OpenVINO-dev-contest, line 368: missing ```infer_queue.wait_all()``` function here

adrianboguszewski: Good point. Thanks

notebooks/109-performance-tricks/109-latency-tricks.ipynb
adrianboguszewski: You're absolutely right. The latency will be the same.

eaidova:
@adrianboguszewski I think it also can be useful to use benchmark app as source for optimal configuration selection (it contains parameter for getting exact configuration used during benchmarking in json format), can we add such section to notebook?

adrianboguszewski:
<p>taking into account this and the above comment, I prefer to go with FP16 only</p>

---

 <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/790/discussion/#comment-1486621416'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

adrianboguszewski:
<p>Added better explanation</p>

---

 <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/790/discussion/#comment-1486636024'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

adrianboguszewski:
<p>That's why I used "usually". I added more explanation why sometimes it's not true.</p>

---

 <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/790/discussion/#comment-1486621475'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

adrianboguszewski:
@eaidova, ready for the next iteration

adrianboguszewski:
<p>That's why we added the following disclaimer: "<strong>NOTE</strong>: Many of the steps presented below will give you better performance. However, some of them may not change anything if they are strongly dependent on either the hardware or the model. Please run this notebook on your computer with your model to learn which of them makes sense in your case."</p><p> </p><p>Therefore, I don't want to skip this trick, as it works for many cases (all I tested).</p>

---

 <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/790/discussion/#comment-1498737370'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

adrianboguszewski:
@paularamo @ryanloney, this is finally ready and we're asking you to review this. I hope you can do it in one week as we want to use it in our new article.

raymondlo84:
ONNX model on CPU: 0.0082 seconds per image (121.28 FPS)

The math looks wrong?

raymondlo84:
image
a bit hard to read, but i guess it's ok but not ideal.

raymondlo84:
image
where is the original diagram? somehow this looks very blurry

raymondlo84:
we should reupload this.

raymondlo84 (changes requested):
a few issues here and there, I think the first inference is always slower. We shall highlight that instead of mixing all the average in the final results.

raymondlo84:
Looks cool. I think we should use 'ms' in the final graph to make it look cleaner. Like 0.00100 is hard to understand. 

Also, we should print out the 'first' inference latency. It's a piece of useful information rather than 'just ignore'.

raymondlo84:
I will merge it if no more updates. I think it's going to be fun to see this running on more hardware and make a little benchmark database.

adrianboguszewski:
@raymondlo84, oh yeah I forgot to print the first inference time - fixed now

yuxu42:
@xczhai Could you please take a review from CPU perspective? Thanks!

xczhai:
- I can't understand why `OpenVINO IR model + more inference threads` is better than `OpenVINO IR model in latency mode`. From the `OpenVINO IR model in latency mode`, the model runs on `CPU`, right? If so, it is better to add  `OpenVINO IR model in latency mode + CPU in latency mode` to verify whether the overhead exists between `OpenVINO IR model in latency mode` though it is a minimal fine-tuning check.
- From the precision view, as Xu Yu mentioned, `ov_model = mo.convert_model(onnx_path, compress_to_fp16=True)` doesn't work for i7-12700 CPU because the native hardware doesn't support fp16 for now. If you check the property of `INFERENCE_PRECISION_HINT`, you will find the precision is fp32 instead of fp16. `precision = compiled_model.get_property("INFERENCE_PRECISION_HINT")`. It is better to clarify this part.

adrianboguszewski:
> * I can't understand why `OpenVINO IR model + more inference threads` is better than `OpenVINO IR model in latency mode`. From the `OpenVINO IR model in latency mode`, the model runs on `CPU`, right? If so, it is better to add  `OpenVINO IR model  + CPU in latency mode` to verify whether the overhead exists between `OpenVINO IR model in latency mode` though it is a minimal fine-tuning check.
> * From the precision view, as Xu Yu mentioned, `ov_model = mo.convert_model(onnx_path, compress_to_fp16=True)` doesn't work for i7-12700 CPU because the native hardware doesn't support fp16 for now. If you check the property of `INFERENCE_PRECISION_HINT`, you will find the precision is fp32 instead of fp16. `precision = compiled_model.get_property("INFERENCE_PRECISION_HINT")`. It is better to clarify this part.

* The default number of inference threads is lower than the number of available logical cores. That's why increasing the number of infer threads give you better latency than using LATENCY mode (with the default number of threads). I already filled the bug report (103649) about the default number of threads as it seems to be set incorrectly.
* Yes. I updated the statement.

raymondlo84 (approved):
Go ahead. Looking good ;)

raymondlo84:
> Yes, I know that `Latency` does not hit the lowest latency value since it follows an `empirical formula`. From the test result view, if you know the best device, just choose it instead of `AUTO`. So I think there is still some space to improve. e.g.
> 
> * `OpenVINO IR model + CPU + shared memory in latency mode`
> * `OpenVINO IR model + CPU + shared memory + more inference threads`.
> 
> I find the above scenarios can improve the latency a bit. Take a try if it's not too late.



> @xczhai, as we agreed, the performance improvement in the above scenarios fluctuates so we decided not to add these scenarios

Would be good to add these to the notes in the notebooks so information won't be loss. Otherwise all good and we will do another round of PR if we find anything interesting.

adrianboguszewski:
@OpenVINO-dev-contest, our plan for the next notebook is the throughput mode. So if the multiple-model topic is interesting to you, feel free to implement that notebook yourself :)

PR #805, opened 2023-02-02
225: add VAE encoder conversion

eaidova:


raymondlo84:
there was a typo here?

        def forward(self, image):
            return e.encode(image).latent_dist.sample()

e?? 

-->

        def forward(self, image):
            return self.vae.encode(image).latent_dist.sample()

raymondlo84:
We should change the "CPU" -> "AUTO" as well.

raymondlo84:
https://mpost.io/best-100-stable-diffusion-prompts-the-most-beautiful-ai-text-to-image-prompts/

Also, we may wanna try a more interesting prompt? I think the current example is cool, but not cool enough :)

eaidova:
> there was a typo here?
> 
> ```
>     def forward(self, image):
>         return e.encode(image).latent_dist.sample()
> ```
> 
> e??
> 
> -->
> 
> ```
>     def forward(self, image):
>         return self.vae.encode(image).latent_dist.sample()
> ```

yes, thanks for note, fixed

eaidova:
> We should change the "CPU" -> "AUTO" as well.

changed

eaidova:
> https://mpost.io/best-100-stable-diffusion-prompts-the-most-beautiful-ai-text-to-image-prompts/
> 
> Also, we may wanna try a more interesting prompt? I think the current example is cool, but not cool enough :)

please select what you wish, I'll update result

raymondlo84:
image

Any idea on this? 

https://github.com/bes-dev/stable_diffusion.openvino

https://huggingface.co/raymondlo84/stable-diffusion-v1-4-openvino-fp16

I get this when I run it with our converted models.

 python demo.py --prompt "Street-art painting of Emilia Clarke in style of Banksy, photorealism" --model raymondlo84/stable-diffusion-v1-4-openvino-fp16

eaidova:
probably, I need to align input and output names used in demo

raymondlo84:
Awesome, and if this works, we can provide others the model and ask the other repos to either update the instructions or change the repos, or update the demo code. etc...

https://huggingface.co/raymondlo84/stable-diffusion-v1-4-openvino-fp16/tree/main

eaidova:
@raymondlo84 does it planned to use only for text to image or other scenarios too?
I have some difference in how encoder was converted, so possibly if it is planned to use I need to update the conversion function

raymondlo84:
> @raymondlo84 does it planned to use only for text to image or other scenarios too? I have some difference in how encoder was converted, so possibly if it is planned to use I need to update the conversion function

I like better code for long run. I think this is one of the most asked questions (how to convert) and so let's make it good?

raymondlo84:
input_names=['input_ids'],  # model input name for onnx representation
                output_names=['last_hidden_state', 'pooler_out'],  # model 

Also, this is the issue right? We hard coding these names... Is there a better way?

eaidova:
> > @raymondlo84 does it planned to use only for text to image or other scenarios too? I have some difference in how encoder was converted, so possibly if it is planned to use I need to update the conversion function
> 
> I like better code for long run. I think this is one of the most asked questions (how to convert) and so let's make it good?

I updated model input names and encoder part, please reconvert models

eaidova:
> ```
>             input_names=['input_ids'],  # model input name for onnx representation
>             output_names=['last_hidden_state', 'pooler_out'],  # model 
> ```
> 
> Also, this is the issue right? We hard coding these names... Is there a better way?

if we do not hardcode them, they will be assigned default from model, usually it is just set of numbers, which can be different from run to run, so it is better just align names with used in demo (in our case we do not use names at all, so no difference)

raymondlo84:
> > ```
> >             input_names=['input_ids'],  # model input name for onnx representation
> >             output_names=['last_hidden_state', 'pooler_out'],  # model 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Also, this is the issue right? We hard coding these names... Is there a better way?
> 
> if we do not hardcode them, they will be assigned default from model, usually it is just set of numbers, which can be different from run to run, so it is better just align names with used in demo (in our case we do not use names at all, so no difference)

oh interesting. that's why we do the property and listing the model info... like dump info etc

raymondlo84:
> > > @raymondlo84 does it planned to use only for text to image or other scenarios too? I have some difference in how encoder was converted, so possibly if it is planned to use I need to update the conversion function
> > 
> > 
> > I like better code for long run. I think this is one of the most asked questions (how to convert) and so let's make it good?
> 
> I updated model input names and encoder part, please reconvert models

let me update and let you know... EH... gonna do the whole upload again... D:

raymondlo84:
oh, it only updated the xml file :D that's so cool

raymondlo84:
almost there... now the UNET :

raymondlo84:
image

raymondlo84:
Ok, I think I forgot to delete the UNET folder... that's why. Let me fix it.

raymondlo84:
> > almost there... now the UNET :
> 
> are you sure that you remove directory unet before running conversion? Looks like you still use previous onnx model. in notebook code, I have the same input name

smart smart smart smart ...

raymondlo84:
image

raymondlo84:
The results looks off, but well~ :D it ran! 

output

raymondlo84:
Street-art painting of Emilia Clarke in style of Banksy, photorealism (if we run this as our prompt, we get a bit of a crazy result too).

I think now the next step is to do a side by side compare of the demo code! :)

eaidova:
> The results looks off, but well~ :D it ran!
> 
> output

we use fine-tuned version openjorney, while in repo used original sd (due to legal issues), but I have it from initial experiments, we can align models

raymondlo84:
> we use fine-tuned version openjorney, while in repo used original sd (due to legal issues), but I have it from initial experiments, we can align models

=D awesome!

raymondlo84:
https://github.com/bes-dev/stable_diffusion.openvino/tree/master 

feel free to test it too, here is the prompt I used.

python demo.py --prompt "Street-art painting of Emilia Clarke in style of Banksy, photorealism" --model raymondlo84/stable-diffusion-v1-4-openvino-fp16

eaidova:
another reason why you get different result, is the seed for the random generator. as I can  see, it is not pinned in demo if not provided, just used it fully randomly

raymondlo84:
got it. also, we need bigger steps like 20+ or the result will look horrible.

Can we cross check the demo btw? The results always look like a shadow, and it seems the weight or similar functions are off

raymondlo84:
50
image
20
image
10
image

raymondlo84:
output

and ran it again with the seed, and the 'grey' out may have to do with how to scale/weight in the model flow.

eaidova:
it is very sad that demo uses too old diffusers lib :( At least, it makes it incompatible with new models like sd 2.0, our notebook is more friendly to that. As well, as some more convenient converted model (before alignment, I reduces pre/postprocessing code and align it with original implementation)

raymondlo84:
> it is very sad that demo uses too old diffusers lib :( At least, it makes it incompatible with new models like sd 2.0, our notebook is more friendly to that. As well, as some more convenient converted model (before alignment, I reduces pre/postprocessing code and align it with original implementation)

Let's build a v2 version of that repos then. :) We'll follow up on this again.

eaidova:
I still trying download model from hf hub, I have some suspicious why quality is poor in demo setup, trying to verify it

eaidova:
output
 My assamptiomn was correct, I n eed update vae_decoder export (Hope it is the last change :))

raymondlo84:
> I n eed update vae_decoder export (Hope it is the last change :))

Also, we shall update the prompt to show a beautiful pic too =) You can pick your favour one :) and then we can update our readme (the main page) one.

eaidova:
I think some landscape will be excellent 
result

eaidova:
@raymondlo84 please reexport vae_decoder according the last notebook version. Now it can work in both demo and notebook with the same quality

adrianboguszewski (commented):
I deleted old models and it dies every time at this step:
image

eaidova:
<p>fixed</p>

---

 <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/805/discussion/#comment-1415739314'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

eaidova:
> I deleted old models and it dies every time at this step
probably it is due to storing models in memory after mo.convert_model, I remove this step, please try again

adrianboguszewski:
> > I deleted old models and it dies every time at this step
> 
> probably it is due to storing models in memory after mo.convert_model, I remove this step, please try again

Still the same. My 32GB of RAM is too little. This is not good. I don't expect an average user would have more

adrianboguszewski:
I will try once again in the next week

raymondlo84:
neat. I was reading the decoder export code yesterday night too! :D I saw some random code there, and now it got cleaned up 📦 

Testing it now and will post my results next.

raymondlo84:
WOW IT WORKED!! 
output
image
And that's LOT AND LOT faster... like 6.6it/s!! D: vs 1.x...!!
image
image

```
python demo.py --prompt "Street-art painting of Emilia Clarke in style of Banksy, photorealism" --model raymondlo84/stable-diffusion-v1-4-openvino-fp16
```
output

Nice, and all we need to do now is to provide the new repos and no other code changes. :)

Issue #799, opened 2023-01-27
Merge paddle classification notebook into paddle conversion notebook


adrianboguszewski:
The 103 and 214 notebooks are very similar. It would be beneficial to merge them into one -> 103 with all features shown.

adrianboguszewski:
@OpenVINO-dev-contest do you think you can merge them?

OpenVINO-dev-contest:
214 was created for 2022.1 launch, it's a bit out-of-data now, but the features in it can be spited and reused into other notebook topics. The features include: 
[Preprocessing API](https://docs.openvino.ai/latest/openvino_docs_OV_Runtime_UG_Preprocessing_Overview.html)
Directly Loading a PaddlePaddle Model
[Auto-Device Plugin](https://docs.openvino.ai/latest/openvino_docs_IE_DG_supported_plugins_AUTO.html)
[AsyncInferQueue Python API](https://docs.openvino.ai/latest/api/ie_python_api/_autosummary/openvino.runtime.AsyncInferQueue.html?highlight=asyncinferqueue#openvino.runtime.AsyncInferQueue)
[Performance Hints](https://docs.openvino.ai/nightly/openvino_docs_OV_UG_Performance_Hints.html)
LATENCY Mode
THROUGHPUT Mode

My idea is merge ```paddle directly loading``` into 103, and move ```Preprocessing API``` to another one.

adrianboguszewski:
That sounds good to me. @eaidova, any of your thoughts?

eaidova:
@sbalandi working on preprocessing api tutorial, probably it is better to sync with her
I added direct paddle loading in 002 (like all other model formats), but it is also good to have it in 103 too

OpenVINO-dev-contest:
If ```preprocessing api``` and  ```paddle directly loading``` are already included, then i think we can remove 214.

adrianboguszewski:
@OpenVINO-dev-contest could you create a PR to remove 214, then?

PR #807, opened 2023-02-02
remove 214

OpenVINO-dev-contest:
https://github.com/openvinotoolkit/openvino_notebooks/issues/799


adrianboguszewski (changes requested):
Chinese readme also?

eaidova (approved):

PR #810, opened 2023-02-03
update yolov8 with segmentation support

eaidova:


adrianboguszewski (changes requested):
Wow! The outcome is great! However, I recommend to look at the following:
- there are some typos and grammar issues, and it's difficult to point out all of them in either github or reviewNB - I can recommend using a free account of the Grammarly plugin, which can help you to spot all those issues
- this notebook's running time was too long and now it's even longer - I recommend decreasing the time of benchmark_app to 15s and making accuracy checking optional (or place it at the end) as it's the longest section of the notebook very difficult to skip (stop kernel and re-run cells below and then again)

PR #825, opened 2023-02-10
Add 232 CLIP Language Saliency Map

apaniukov:
This notebook shows how to build a saliency map from a text query using CLIP

* Explanation of a _saliency map_ and how it can be used.
* Overview of the CLIP neural network and its usage in generating saliency maps.
* How to split a neural network into parts for separate inference.
* How to speed up inference with OpenVINO™ and asynchronous execution.


MaximProshin:
@apaniukov , can you please replace POT by NNCF in further steps as we want to recommend NNCF as the default compression tool since next OV release.

adrianboguszewski:
@apaniukov please change the number to 214

eaidova:
> @apaniukov please change the number to 214

I think for consistency, it is better to have it as 232, because we have 228 with clip model usage, this step is more advanced (and also not always perform good), so I prefer to keep them closer

eaidova:
> @apaniukov , can you please replace POT by NNCF in further steps as we want to recommend NNCF as the default compression tool since next OV release.

I replaced mention in next pr

PR #826, opened 2023-02-13
117 construct model dir

OpenVINO-dev-contest:
demonstrate how to construct model dir


OpenVINO-dev-contest:
> @OpenVINO-dev-contest we got ConnectionError for 117 notebook on macos. Could you please take a look, is it expected?(maybe we need describe this limitation in readme or notebook?)
Hi, I think this error is caused by that the port 9000 is in use in MacOS CI, so I have put a comments after the command :```If the serving port 9000 is already in use, please switch it to another available port on your system.```

OpenVINO-dev-contest:
> @OpenVINO-dev-contest we got ConnectionError for 117 notebook on macos. Could you please take a look, is it expected?(maybe we need describe this limitation in readme or notebook?)

Hi, I think this error is caused by that the port 9000 is in use in MacOS CI, so I have put a comments after the command :```If the serving port 9000 is already in use, please switch it to another available port on your system.```

eaidova:
@raymondlo84 @adrianboguszewski  can you please help with checking work of 117 notebook on macos?
Just for understanding that it is our ci issue or general problem and can be hanlered in code or something else

OpenVINO-dev-contest:
> I got an exception image

Thanks for your testing, does it happen during package import ?

OpenVINO-dev-contest:
截屏2023-02-16 23 13 07
I tested it on my Intel Macbook, the results are all correct

adrianboguszewski:
@OpenVINO-dev-contest It happens when importing ovms_client. My mac has an M1 inside.

OpenVINO-dev-contest:
> @OpenVINO-dev-contest It happens when importing ovms_client. My mac has an M1 inside.

I think it may caused by a wrong version of Tensorflow, could you remove the current version and reinstall it with ```pip install ovmsclient```

adrianboguszewski:
> > @OpenVINO-dev-contest It happens when importing ovms_client. My mac has an M1 inside.
> 
> I think it may caused by a wrong version of Tensorflow, could you remove the current version and reinstall it with `pip install ovmsclient`

It will not work. There is no standard tensorflow version for mac M1. We have to use tensorflow-macos, which causes the import error above. mac M1 is a lot of pain in the case of OpenVINO...

OpenVINO-dev-contest:
> > > @OpenVINO-dev-contest It happens when importing ovms_client. My mac has an M1 inside.
> > 
> > 
> > I think it may caused by a wrong version of Tensorflow, could you remove the current version and reinstall it with `pip install ovmsclient`
> 
> It will not work. There is no standard tensorflow version for mac M1. We have to use tensorflow-macos, which causes the import error above. mac M1 is a lot of pain in the case of OpenVINO...

yes thats a big trouble

adrianboguszewski (commented):
What about that message?
image

OpenVINO-dev-contest:
> What about that message?
> image

this is from OVMS team’s suggetion and you can check the orignal PR

adrianboguszewski:
> this is from OVMS team’s suggestion and you can check the conversation in orignal PR

`docker rm` doesn't have any effect here as you have a `--rm` flag when running. I think you should use either `--rm flag` or `docker rm ovms`

adrianboguszewski (approved):

eaidova (approved):

PR #828, opened 2023-02-13
Remove omz tools from 401 and 402

adrianboguszewski:
Ticket-id: *103132*

### DIFF
--- a/notebooks/401-object-detection-webcam/401-object-detection.ipynb
+++ b/notebooks/401-object-detection-webcam/401-object-detection.ipynb
@@ -21,22 +21,20 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "pycharm": {
-     "name": "#%%\n"
-    }
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "import collections\n",
-    "import os\n",
     "import sys\n",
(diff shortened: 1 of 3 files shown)


notebooks/401-object-detection-webcam/401-object-detection.ipynb
```diff
@@ -21,22 +21,20 @@
   {
```
eaidova, line 21: <pre class="ql-syntax" spellcheck="false">Line #12.                          f"--compress_to_fp16 {True if precision == 'FP16' else False} " \
</pre><p>not sure that it is work, probably you should be:</p><p>if precison == "FP16":</p><p>convert_command + " -- compress_to_fp16"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/828/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

adrianboguszewski: Tested. It works as expected

eaidova (approved):

ryanloney:
Great to see this change! Thank you.

PR #844, opened 2023-02-19
notebooks conversion

eaidova:
Uodate for noteboks conversion script.

* introduced skipping for notebooks required additional conditions (more memory, GPU e.t.c.), output of notebooks will be stored in the repo and conversion will use it instead execution on machine
* added timeout for killing run if notebook hangs
* better reporting of failures

TO DO:

- [ ]  make sure that all ignored notebooks have output
- [ ] fix failures

### DIFF
--- a/.ci/convert_notebooks.py
+++ b/.ci/convert_notebooks.py
@@ -1,0 +1,88 @@
+import argparse
+from pathlib import Path
+import time
+import subprocess
+import shutil
+
+
+def arguments():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--exclude_execution_file")
+    parser.add_argument("--timeout", type=float, default=3600,
+                        help="timeout for notebook execution")
+    parser.add_argument("--rst_dir", type=Path,
+                        help="rst files output directory", default=Path("rst"))
+    parser.add_argument("--html_dir", type=Path,
+                        help="html files output directory", default=Path("html"))
+    parser.add_argument("--markdown_dir", type=Path,
+                        help="markdown files output directory", default=Path("markdown"))
+
+    return parser.parse_args()
+
+
+def prepare_ignore_execution_list(input_file):
+    with Path(input_file).open("r") as f:
+        lines = f.readlines()
+    return list(map(str.strip, lines))
+
+
+def main():
+    args = arguments()
+    ignore_execution_list = []
+    failed_notebooks = []
+    markdown_failed = []
+    rst_failed = []
+    html_failed = []
+    if args.exclude_execution_file is not None:
+        ignore_execution_list = prepare_ignore_execution_list(
+            args.exclude_execution_file)
+    ROOT = Path(__file__).parents[1]
+    notebooks_dir = ROOT / "notebooks"
+    notebooks = sorted(list(notebooks_dir.rglob('**/*.ipynb')))
+    for notebook in notebooks:
+        notebook_path = notebook.relative_to(ROOT)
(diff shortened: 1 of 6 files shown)


adrianboguszewski (commented):
The only thing I would change is the name of the txt file. Maybe "exclude_notebooks_from_conversion.txt" is more meaningful as we have many ignore lists for different processes?

adrianboguszewski:
And I also discovered locally the exclude_run.txt is ignored. 
image

.ci/convert_notebooks.py
```diff
@@ -0,0 +1,88 @@
...
+                continue
+        else:
+            shutil.copyfile(notebook_path, notebook_executed)
+        markdown_retcode = subprocess.run(["jupyter", "nbconvert", "--to", "markdown", str(notebook_executed), "--output-dir", str(args.markdown_dir),
+                                          "--TagRemovePreprocessor.remove_all_outputs_tags=hide_output --TagRemovePreprocessor.enabled=True"], timeout=args.timeout).returncode
+
+        html_retcode = subprocess.run(["jupyter", "nbconvert", "--to", "html", str(notebook_executed), "--output-dir", str(args.html_dir),
+                                          "--TagRemovePreprocessor.remove_all_outputs_tags=hide_output --TagRemovePreprocessor.enabled=True"], timeout=args.timeout).returncode
+        rst_retcode = subprocess.run(["jupyter", "nbconvert", "--to", "rst", str(notebook_executed), "--output-dir", str(args.rst_dir),
+                                          "--TagRemovePreprocessor.remove_all_outputs_tags=hide_output --TagRemovePreprocessor.enabled=True"], timeout=args.timeout).returncode
```
adrianboguszewski, line 61: Shouldn't be under `if str(notebook_path) not in ignore_execution_list`?

adrianboguszewski (changes requested):

.ci/convert_notebooks.py
eaidova: no, we have some notebooks which store output inside and we no need to spend time to execute them on ci (e.g. notebook requires GPU, but ci machine does not have it), so all actions except execution are acceptable for them, we just avoid execution

eaidova:
> And I also discovered locally the exclude_run.txt is ignored. image

it is ok, in your logs started only conversion of notebook, but there is no line where notebook executed

adrianboguszewski:
I see. What if we want to ignore some notebooks totally, like don't take any pre-generated output, just skip it?

eaidova:
> I see. What if we want to ignore some notebooks totally, like don't take any pre-generated output, just skip it?

I can provide additional parameter for it, but could you please provide example such notebooks?

adrianboguszewski:
I'm thinking about 108, which doesn't have included output, or any notebook that is hanging and also doesn't have included output. I don't say it should be a permanent solution. I just want to have a way to exclude some notebooks temporarily until we fix them as we did in CI checking.

.ci/ignore_convert_full.txt
```diff
@@ -0,0 +1,4 @@
+notebooks/108-gpu-device/108-gpu-device.ipynb
+notebooks/110-ct-segmentation-quantize/data-preparation-ct-scan.ipynb
+notebooks/110-ct-segmentation-quantize/pytorch-monai-training.ipynb
+notebooks/113-image-classification-quantization/113-image-classification-quantization.ipynb
```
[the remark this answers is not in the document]
adrianboguszewski: And 117 doesn't work for me at all. Shall we add it here?

PR #890, opened 2023-03-07
220: fix default mAP score metric incorrect and add FP16 for benchmark

shungfu:
### 895a1b8:
Currently, the default mAP score run is not compatible with the original YOLOv5 code. To address this, I propose modifying the configuration options to ensure consistency between the notebook and the YOLOv5 code.

### a5fbf6f:
This commit adds an FP16 IR model.
Additionally, I suggest converting the OpenVINO IR model using the mo command instead of export. This will improve flexibility to export FP16 and it also fits the workflow of 226-yolov7-optimization.

### 6181fc4:
I have added a benchmark for FP32/FP16/INT8 for both CPU and GPU in this commit. This will provide valuable insights into the performance of the code on different hardware configurations.


notebooks/220-yolov5-accuracy-check-and-quantization/220-yolov5-accuracy-check-and-quantization.ipynb
```diff
@@ -46,11 +46,13 @@
     "import matplotlib.pyplot as plt\n",
```
eaidova, line 46: <pre class="ql-syntax" spellcheck="false">Line #38.    optimized_save_dir = Path(Path.cwd()).joinpath(config["model_int8"]["save_path"])
</pre><p>probably </p><p>Path(config["model_int8"]["save_path"])) should be enough</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/890/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

eaidova, line 46: <pre class="ql-syntax" spellcheck="false">Line #1.    # Inference FP32 model (OpenVINO IR)
</pre><p>not all systems has GPU device installed, could you please make benchmarking on GPU optional?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/890/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

eaidova: I mean if "GPU' in core.available_devices, then run benchmark on GPU or something similar

eaidova:
@shungfu thank you for your update. Also it will be nice to have output inside notebook if it is possible

notebooks/220-yolov5-accuracy-check-and-quantization/220-yolov5-accuracy-check-and-quantization.ipynb
shungfu: Thank you for your suggestion. I agree that using Path(config["model_int8"]["save_path"]) directly would simplify the code.

shungfu: Sure, I will make the GPU optional for the benchmark by modifying the code as follows.

```
from openvino.runtime import Core

core = Core()
device = 'GPU' if 'GPU' in core.available_devices else 'CPU'
print(f"Benchmark runnin on Device: {device}")

# Inference FP32 model (OpenVINO IR)
!benchmark_app -m {fp32_path}.xml -d {device} -api async
```

eaidova: this variant does not allow to compare CPU vs GPU performance in the same notebook without code modification, I think it will be better to keep CPU benchmarking part and just put GPU part under condition:

```
print("Inference FP32 model (OpenVINO IR) on CPU")
!benchmark_app -m  {fp32_path}.xml -d CPU -api async
if if 'GPU' in core.available_devices:
    print("Inference FP32 model (OpenVINO IR) on GPU")
    !benchmark_app -m  {fp32_path}.xml -d GPU -api async
```

Also sometimes benchmarking can be very long, -t parameter allow reduce benchmarking time, you can consider to specify it

shungfu: Thank you for your suggestion. I will keep both CPU and GPU but just put GPU part under condition. Also I will add the -t parameter with Time 15 on benchmark_app.

eaidova (approved):

PR #891, opened 2023-03-07
Merging 210 to 110 family

adrianboguszewski:
Keeping 210 in 110 family makes more sense, as it's part of the series. It also allowed removing the utils used by 110 notebooks only.

It also required to use torch hub in 109.


adrianboguszewski:
Tested 211 and 215 locally in a docker container. They work.

eaidova (approved):

PR #928, opened 2023-03-17
Proofreading new notebooks

sgolebiewski-intel:
Proofreading content of the notebooks.


adrianboguszewski (commented):
Thanks for fixing so many language issues :)
I have two questions:
1. Why have relative paths to README been replaced by absolute URLs?
2. Why has "we" been replaced by "you"?

sgolebiewski-intel:
> Thanks for fixing so many language issues :) I have two questions:
> 
> 1. Why have relative paths to README been replaced by absolute URLs?
       First of all, it was incoherent in the current state of the notebooks, and using my best judgement I decided that the absolute link to the main README file would be a better solution, not liable to changes in file structure on the repo.
> 2. Why has "we" been replaced by "you"?
       Again, it was incoherent and there were both pronouns used interchangeably in the notebooks. To make it coherent with our documentation I introduced the use of "you", which in my opinion entails more focus on the user of the notebooks.

notebooks/116-sparsity-optimization/116-sparsity-optimization.ipynb
```diff
@@ -105,11 +108,12 @@
    ]
   },
   {
+   "attachments": {},
    "cell_type": "markdown",
    "id": "227d97eb-e91e-48bb-bba9-30598e49bb4f",
    "metadata": {},
    "source": [
-    "For benchmarking, we will use OpenVINO's benchmark application and let's put the IRs into a single folder."
+    "For benchmarking, you will use OpenVINO's benchmark application. Make sure to put the IRs into a single folder."
```
eaidova, line 116: probably meaning is different. The previous variant speaks that code below put all files in the same directory, current version speaks that the user is responsible for downloading and should check that all files will be downloaded in the same folder.

notebooks/118-optimize-preprocessing/118-optimize-preprocessing.ipynb
```diff
@@ -367,15 +373,15 @@
    "source": [
     "### Preprocessing Steps\n",
     "\n",
-    "Now, the sequence of preprocessing steps can be defined. More information about preprocessing steps, please, see [here](https://docs.openvino.ai/latest/classov_1_1preprocess_1_1PreProcessSteps.html#class-ov-preprocess-preprocesssteps).\n",
+    "Now, the sequence of preprocessing steps can be defined. For more information about preprocessing steps, see the [PreProcessSteps](https://docs.openvino.ai/latest/classov_1_1preprocess_1_1PreProcessSteps.html#class-ov-preprocess-preprocesssteps).\n",
```
eaidova, line 376: not sure that it is good idea to use camel case for link name, as there is no direct usage this class in the code below, it is not obvious that it is class name and how it is connected with common flow. Also python API link will be more suitable than c++
https://docs.openvino.ai/latest/api/ie_python_api/_autosummary/openvino.preprocess.PreProcessSteps.html#openvino.preprocess.PreProcessSteps

notebooks/218-vehicle-detection-and-recognition/218-vehicle-detection-and-recognition.ipynb
```diff
@@ -315,7 +315,7 @@
     "        if conf > threshold:\n",
     "            # Convert float to int and multiply corner position of each box by x and y ratio\n",
     "            # In case that bounding box is found at the top of the image, \n",
-    "            # we position upper box bar little bit lower to make it visible on image \n",
+    "            # you position upper box bar little bit lower to make it visible on image \n",
```
eaidova, line 318: l think it will bw better in general form, without pronounse, it is description how code works, so there is no actor

sgolebiewski-intel, line 318:
```suggestion
    "            # upper box  bar should be positioned little bit lower to make it visible on image \n",
```

notebooks/116-sparsity-optimization/116-sparsity-optimization.ipynb
sgolebiewski-intel, line 116: How about: 

```suggestion
    "For benchmarking, you will use OpenVINO's benchmark application. Use the code below to put the IRs into a single folder."
```

sgolebiewski-intel:
> @sgolebiewski-intel I think it is better to stay relative path on readme for installation. According notebooks branching strategy, when release become published, we create branch for previous official release and users still can use previous package. Now these changes will requires to update all these links with different branch, because there is no guarantee that we will not change installation procedure in one of the next releases. So from repo maintanence side it is looks better to keep relative links.
> 
> Another point, for users who clone repo and open it in IDE, now it will open browser to see installation guide instead of open local file in the same IDE, which also can be less convinient

Ok, will change it to relative path. Commit underway.

adrianboguszewski (changes requested):
Great work! We prefer to use relative paths instead of absolute, as it works better with different branches.

sgolebiewski-intel:
> Great work! We prefer to use relative paths instead of absolute, as it works better with different branches.

Updated with https://github.com/openvinotoolkit/openvino_notebooks/pull/928/commits/48f2f2a3b6dc8085d81bba75c2f09aab4d6f2f73

sawradip:
> Great work! We prefer to use relative paths instead of absolute, as it works better with different branches.

As well as across different forks.

sgolebiewski-intel:
> 2. Why has "we" been replaced by "you"?

To backup the changes I suggested I want to share some information from one of the style guides. I would like you to refer to the following:
**InfoDev Style Guide.pdf** (page 103), found at
https://wiki.ith.intel.com/pages/viewpage.action?spaceKey=CVSDK&title=Writing+and+layout

I hope this elucidates the point of view I have.

notebooks/116-sparsity-optimization/116-sparsity-optimization.ipynb
sgolebiewski-intel: How about: 

```suggestion
    "For benchmarking, you will use OpenVINO's benchmark application. Use the code below to put the IRs into a single folder."
```

eaidova:
@sgolebiewski-intel  @kblaszczak-intel  thank you for sharing info about technical writing guides. I understand your recommendation for alignment and making language more formal. But there are also some points from our side which should be considered:

* From our side, notebooks are considered not as just technical articles and formal descriptions for steps, it has more common with blogs and tutorials, using `we` we want to emphasize our togetherness and union with the reader, it means that notebook authors  accompanying users in their journey to openvino world :)
* Maybe, it is a subjective opinion, but bringing formality in some cases can make the explanation more complicated (we have a wide audience from beginners to experienced technical engineers, so keeping clear and simple explanations is very important)
* We are trying to have storytelling elements to keep user's attention and involvement they in experiments, in some cases, rephrasing with removing `Let's do something` constructions, notebooks can lose connectivity between code cells and its description.

So my question is it possible to relax the rules for notebooks?

kblaszczak-intel:
I'm sure we can work out a solution. For now, let's not mix fixes and tone adjustments :) I suggest we choose one notebook, when we have a bit more time, and use it as a draft. We can see what language fits and satisfies the more relaxed character of tutorials. If we get agreement, then we can move it to other articles.

notebooks/110-ct-segmentation-quantize/README.md
```diff
@@ -10,15 +10,15 @@ This folder contains five notebooks that show how to train,
...
 
-2\. [Train a 2D-UNet Medical Imaging Model with PyTorch Lightning](pytorch-monai-training.ipynb)
+2. [Train a 2D-UNet Medical Imaging Model with PyTorch Lightning](pytorch-monai-training.ipynb)
 
 3a. [Convert and Quantize a UNet Model and Show Live Inference using POT](110-ct-segmentation-quantize.ipynb)
 
 3b. [Convert and Quantize a UNet Model and Show Live Inference using NNCF](110-ct-segmentation-quantize-nncf.ipynb)
 
-4\. [Live Inference and Benchmark CT-scan Data with OpenVINO](110-ct-scan-live-inference.ipynb)
+4. [Live Inference and Benchmark CT-scan Data with OpenVINO](110-ct-scan-live-inference.ipynb)
```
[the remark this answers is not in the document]
sgolebiewski-intel: Fix in progress.

ryanloney (changes requested):
We had brief internal chat and decided the language should not change from "we" to "you" please remove these changes

eaidova (approved):

sgolebiewski-intel:
> We had brief internal chat and decided the language should not change from "we" to "you" please remove these changes

Had reverted the changes with https://github.com/openvinotoolkit/openvino_notebooks/pull/928/commits/cec65a381dc4abda5c76d7ca433be5d861bad76a

adrianboguszewski (approved):
Thanks for fixing so many language issues :)

PR #948, opened 2023-03-24
[CI] Transition to treon for test execution

akashchi:
- Use [`treon`](https://github.com/ReviewNB/treon) for test execution
  - Add `treon` to dependencies  
  - Replace `nbval` usage with `treon`
    - Both for the different OSs and in Docker


.ci/validate_notebooks.py
```diff
@@ -78,16 +78,24 @@ def clean_test_artefacts(before_test_files, after_test_files):
...
 
 
-def run_test(notebook_path, report_dir, collect_reports, root):
-    print(f'RUN {notebook_path.relative_to(root)}')
-    report_file = report_dir / f'{notebook_path.name}_report.xml'
+def run_test(notebook_path, root):
+    print(f'RUN {notebook_path.relative_to(root)}', flush=True)
+    
     with cd(notebook_path):
-        existing_files = sorted(list(notebook_path.rglob("*")))
```
akashchi, line 85: Previously, the `existing_files` was always empty because of the `cd` and still relative `notebook_path`: there are no notebook folder in the notebook folder.

eaidova (approved):

PR #971, opened 2023-03-30
Stable Diffusion v2 (quick fixes) 

raymondlo84:
fixed a few typos and changed the final output to use only 3 seconds …


raymondlo84:
> View / edit / reply to [this conversation](https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/971/discussion/#comment-1491200536) on ReviewNB
> 
> _OpenVINO-dev-contest commented on 2023-03-31T02:28:31Z_ ----------------------------------------------------------------
> 
> Line #5.    tokenizer_inpaint = CLIPTokenizer.from_pretrained('openai/clip-vit-large-patch14')
> Hi @eaidova I think the tokenizer has been initialized by `StableDiffusionPipeline ` and we can call it from attributes of `pipe ` directly, so why should we create it object again?

I think the bigger issue I was trying the code is re-running it without knowing the entire flow very carefully as we had shared variable in the two demos.  I renamed them as much as we can and so at least we can re-run any part of the demo block by block now. (kinda...). I think we should avoid 'reusing' variable or rewriting in general? What do you think?

OpenVINO-dev-contest:
> > View / edit / reply to [this conversation](https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/971/discussion/#comment-1491200536) on ReviewNB
> > _OpenVINO-dev-contest commented on 2023-03-31T02:28:31Z_ ----------------------------------------------------------------
> > Line #5.    tokenizer_inpaint = CLIPTokenizer.from_pretrained('openai/clip-vit-large-patch14')
> > Hi @eaidova I think the tokenizer has been initialized by `StableDiffusionPipeline ` and we can call it from attributes of `pipe ` directly, so why should we create it object again?
> 
> I think the bigger challenge was "re-running" without knowing the entire flow very carefully as we had shared variable in the two demos. I renamed the variables as much as we can and so at least we can re-run any part of the demo block by block now. (kinda...). I think we should avoid 'reusing' variable or rewriting in general? What do we think?

To be honest, I prefer the implementation method with [optimum-intel](https://github.com/huggingface/optimum-intel), cuz its simpler to play with, but you can't see the full process behind it .

eaidova:
<p>to make pipeline independent from diffusers</p>

---

 <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/971/discussion/#comment-1491200536'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

eaidova:
> > > View / edit / reply to [this conversation](https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/971/discussion/#comment-1491200536) on ReviewNB
> > > _OpenVINO-dev-contest commented on 2023-03-31T02:28:31Z_ ----------------------------------------------------------------
> > > Line #5.    tokenizer_inpaint = CLIPTokenizer.from_pretrained('openai/clip-vit-large-patch14')
> > > Hi @eaidova I think the tokenizer has been initialized by `StableDiffusionPipeline ` and we can call it from attributes of `pipe ` directly, so why should we create it object again?
> > 
> > 
> > I think the bigger challenge was "re-running" without knowing the entire flow very carefully as we had shared variable in the two demos. I renamed the variables as much as we can and so at least we can re-run any part of the demo block by block now. (kinda...). I think we should avoid 'reusing' variable or rewriting in general? What do we think?
> 
> To be honest, I prefer the implementation method with [optimum-intel](https://github.com/huggingface/optimum-intel), cuz its simpler to play with, but you can't see the full process behind it .

the limitations of optimum pipeline are that it export and run models with dynamic shapes which does not work on GPU, the second point that it still will not work with https://github.com/bes-dev/stable_diffusion.openvino, but our version can

here my flork which enables stable diffusion v2 in demo https://github.com/eaidova/stable_diffusion.openvino/tree/sd2 (also allow run model stored locally without uploading hf hub)

raymondlo84:
> View / edit / reply to [this conversation](https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/971/discussion/#comment-1491302684) on ReviewNB
> 
> _eaidova commented on 2023-03-31T05:06:05Z_ ----------------------------------------------------------------
> 
> my opnion that additional headers no need here and above because in general structure all first part related to Text-to-Image showcase, it looks like iwriting book you will start each chapter from book header
> 
> but I do not insist

Ok-- agree. And patched and now the sessions look cleaner and all aligned.

eaidova (approved):

PR #999, opened 2023-04-10
fix bandit scan issues

eaidova:


eaidova:
@akashchi I got 

Run github/issue-labeler@v3.1
Could not add label(s) fix to issue #999
Error: AggregateErr

is it expected? I'm not sure that this strategy will works, because majority internally detected issues can be submitted via jira and have no related github issues

PR #1018, opened 2023-04-20
Fix yolov8 validator initialization with args

sammysun0711:
Problem: 
According to [ultralytics 8.0.43](https://github.com/ultralytics/ultralytics/blob/d42af1363fcaf48089b18ea63e0e0ad31722e448/ultralytics/yolo/engine/validator.py#L59) BaseValidator initialize as follow: 
```
def __init__(self, dataloader=None, save_dir=None, pbar=None, logger=None, args=None)
```
But in notebook, det_validator is initialized as follow: 
```
det_validator = det_model.ValidatorClass(args)
```
the args was pass to first parameter `dataloader` instead of `args`, which lead to problem that we can not make modified args effective for validator & dataloader required for NNCF PTQ: 

Simple reproduce, say we want data loader to generate image size with 320x320 instead of default 640x640: 
```
args = get_cfg(cfg=DEFAULT_CFG)
args.imgsz=320
det_validator = det_model.ValidatorClass(args)
print(det_validator.args.imgsz)
```
det_validator.args.imgsz will be still 640

Solution
```
args = get_cfg(cfg=DEFAULT_CFG)
args.imgsz=320
det_validator = det_model.ValidatorClass(args=args)
print(det_validator.args.imgsz)
```
Now, det_validator.args.imgsz will be 320


adrianboguszewski:
Good catch. Thanks for the fix!

PR #1021, opened 2023-04-24
Move model checkpoints to cloud

igor-davidyuk:
# Prune the repo to remove model checkpoints and data samples.
### Processed
001, 002, 003, 004, 109, 110, 201, 202, 205, 219, 225

### Here is a list of examples with isolated requirements files
- 212-pyannote-speaker-diarization
  - requirements.txt
- 234-encodec-audio-compression
  - requirements.txt


raymondlo84:
how do we upload to the 'cloud' next time? Is there a way we can get access to that so we don't have to refactor again in the future?

igor-davidyuk:
> how do we upload to the 'cloud' next time? Is there a way we can get access to that so we don't have to refactor again in the future?

From my understanding, most data samples and model checkpoints are already available on the web.
However, when we perform pretraining ourselves or have a unique data piece, we should store them in our S3 bucket and fetch them from notebook example code, instead of keeping them in the repo.
@eaidova, could you please clarify the way to do it?

eaidova:
@raymondlo84  @igor-davidyuk  probably it is not for public ears :), I'll answer in internal conversation

eaidova (approved):

PR #1037, opened 2023-05-05
Enhancing Segment Anything Notebook

jacobmarks:
I revamped the notebook to make the workflows more streamlined, reduce duplicated code, and instead of drawing points and masks on images, they are now toggle-able labels.

Here is a summary of changes made:

- Added FiftyOne to requirements.txt for management, visualization, and control over computer vision datasets
- Improved grammar and language in SAM notebook
- Removed matplotlib dependence (including `show_mask()`, `show_points()`, and `show_box()` functions), and replaced all visuals with images from a session of the FiftyOne App. This is more flexible, generalizable to entire datasets, and keeps the focus of the notebook on OpenVINO
- Turned repeated instances of blocks like:

```
coord = np.concatenate([input_point, np.array([[0.0, 0.0]])], axis=0)[None, :, :]
label = np.concatenate([input_label, np.array([-1])], axis=0)[None, :].astype(np.float32)
coord = resizer.apply_coords(coord, image.shape[:2]).astype(np.float32)

inputs = {
    "image_embeddings": image_embeddings,
    "point_coords": coord,
    "point_labels": label,
}

results = ov_predictor(inputs)

masks = results[ov_predictor.output(0)]
masks = postprocess_masks(masks, image.shape[:-1])
masks = masks > 0.0
```

into a function, `generate_mask()`.


jacobmarks:
Hi @paularamo , thanks for taking a look! It seems like the issue might be that you created the dataset with that name, and then tried to create it again with the same name. It is persistent so this re-run will result in an error unless you pass in `overwrite=True`. 

If you are having trouble with naming, you can just create the dataset without a name and a name will be auto-assigned:

```
dataset = fo.Dataset(persistent=True)
```

You can also set the name after creating:

```
dataset.name = "my-name"
```

Let me know if this resolves the issue :)

paularamo:
> Hi @paularamo , thanks for taking a look! It seems like the issue might be that you created the dataset with that name, and then tried to create it again with the same name. It is persistent so this re-run will result in an error unless you pass in `overwrite=True`.
> 
> If you are having trouble with naming, you can just create the dataset without a name and a name will be auto-assigned:
> 
> ```
> dataset = fo.Dataset(persistent=True)
> ```
> 
> You can also set the name after creating:
> 
> ```
> dataset.name = "my-name"
> ```
> 
> Let me know if this resolves the issue :)

Thanks @jacobmarks. I already made your suggestions, another error is popping up. "filepath" is not defined.

image

If I am doing something wrong, let me know, maybe we can also document that in the notebook.

Thanks again :)

paularamo:
Thanks for the fixes :)

paularamo:
> Hi @paularamo , thanks for taking a look! It seems like the issue might be that you created the dataset with that name, and then tried to create it again with the same name. It is persistent so this re-run will result in an error unless you pass in `overwrite=True`.
> 
> If you are having trouble with naming, you can just create the dataset without a name and a name will be auto-assigned:
> 
> ```
> dataset = fo.Dataset(persistent=True)
> ```
> 
> You can also set the name after creating:
> 
> ```
> dataset.name = "my-name"
> ```
> 
> Let me know if this resolves the issue :)

Could you please add `overwrite=true` to avoid further issues in case users run this cell multiple times? We can have that as default cell in the notebook and show them how they could remove the overwrite part with an example after that cell.

paularamo:
> Hi @paularamo , thanks for taking a look! It seems like the issue might be that you created the dataset with that name, and then tried to create it again with the same name. It is persistent so this re-run will result in an error unless you pass in `overwrite=True`.
> 
> If you are having trouble with naming, you can just create the dataset without a name and a name will be auto-assigned:
> 
> ```
> dataset = fo.Dataset(persistent=True)
> ```
> 
> You can also set the name after creating:
> 
> ```
> dataset.name = "my-name"
> ```
> 
> Let me know if this resolves the issue :)

Could you please add `overwrite=true` to avoid further issues in case users run this cell multiple times? We can have that as default cell in the notebook and show them how they could remove the overwrite part with an example after that cell.

paularamo:
Gradio is working again in my environment. Are you using MAC?

Regarding "image" definition we need to keep the variable in the notebook. Please add this where we are reading the file.

`image = cv2.imread('truck.jpg')
image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)`

to avoid this issue: 

image

paularamo:
Thanks for your adding my suggestions. @jacobmarks

paularamo (approved):
@jacobmarks, Thanks for this great contribution to our repo. We love to see how this repository could help more and more devs around the world.

raymondlo84:
We cannot upload all these png files to the repos, that's over 40MB.

jacobmarks:
Hi @adrianboguszewski, thanks for the review. Looks like you may need to run something like the following to set up MongoDB on your computer prior to using FiftyOne (this is a one time thing, and is only necessary on some computers):

```
DB_DIR=~/db
mkdir -p $DB_DIR
mongod --dbpath $DB_DIR --port 12345
```

adrianboguszewski:
Still unsuccessful. The same cell.
image

Is MongoDB required to run the notebook? If so, please add a short information with the link to the instructions on installing mongo.

requirements.txt
```diff
@@ -31,6 +31,7 @@ ipython>=7.16.3
 
 
 # others
+fiftyone>=0.19.0
```
adrianboguszewski, line 34:
```suggestion
```

adrianboguszewski: The installation is done in the notebook

jacobmarks:
Hi @adrianboguszewski, thanks for the additional review. Sorry you're running into this issue. Are you by any chance on ubuntu? 

If so, then one way to circumvent this issue is to install your own version of MongoDB and configure a connection to it. We have documentation [here](https://docs.voxel51.com/user_guide/config.html#configuring-mongodb-connection). In fact, import times will be much faster with this setup.

I also wanted to check in about having FiftyOne in the OpenVINO Notebooks `requirements.txt` file - I think having FiftyOne in the `requirements.txt` will make future collaboration between FiftyOne and OpenVINO on code, content and community that much easier.

adrianboguszewski:
@jacobmarks, we used to have all dependencies in the requirement file, but we had to abandon that approach as we got into dependency hell and had a very long installation time. So now, we keep there dependencies which are required by many notebooks only. If fiftyone becomes that case, we will be happy to add it to the requirements file :)

I'll test mongo solution later.

jacobmarks:
@adrianboguszewski @paularamo 
Updates:

- Merged in the latest from `main`, including the updated `segment-anything` notebook
- Moved `fiftyone` out of `requirements.txt`
- Added code for installing fiftyone on Ubuntu 22
- Changed the name of this notebook to `237-segment-anything-fiftyone.ipynb`
- Added breadcrumb for this notebook to the repo's main README file
- Updated the `237-segment-anything` dir's README in accordance with the 2-notebook contents
- Refactored the notebook, removing `gradio`, and focusing more on explanation
- Added a section to the end showing how to apply OpenVINO SAM to a full dataset with fiftyone.

raymondlo84 (commented):
would also be interested in seeing the RAM usage with the new tools. Let's go with CI and see if we can catch any issues next.

PR #1040, opened 2023-05-09
Quality control AI Recipe [Anomalib - Use case]

paularamo:
Here you will find a new AI recipe working directly with Anomalib, our DL library, for training, testing, and deployment of anomaly detection models.

This recipe should be within the OpenVINO Notebooks, not just for the new content we are preparing with SMG and marketing teams, also because it will help to our customer to make and easy integration with OpenVINO + Anomalib in just one single place.


adrianboguszewski:
Is this recipe standalone? Does it need any code from the OpenVINO notebooks repo?

PR #1053, opened 2023-05-15
110 fix flow

aleksandr-mokrov:
ONNX is excluded. Converting to IR model is before quantization. Converting quantized model doesn't work in this case ignored_scope is committed due to error.

### DIFF
--- a/notebooks/110-ct-segmentation-quantize/110-ct-segmentation-quantize-nncf.ipynb
+++ b/notebooks/110-ct-segmentation-quantize/110-ct-segmentation-quantize-nncf.ipynb
@@ -56,11 +56,7 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "tags": [],
-    "ExecuteTime": {
-     "start_time": "2023-05-15T22:58:00.175956Z",
-     "end_time": "2023-05-15T22:58:00.238621Z"
-    }
+    "tags": []
    },
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


eaidova:
@aleksandr-mokrov I think your changes not what expected. Initially notebook demonstrated quantization using pytorch, now you change it to quantize openvino IR. the root cause of bug described in ticket that pytorch fe does not support direct loading quantized models, you need to just provide path to int8_onnx instead of model to convert_model

notebooks/110-ct-segmentation-quantize/110-ct-segmentation-quantize-nncf.ipynb
```diff
@@ -477,22 +481,24 @@
...
-    "data_loader = torch.utils.data.DataLoader(dataset, batch_size=4)\n",
+    "data_loader = torch.utils.data.DataLoader(dataset)\n",
     "calibration_dataset = nncf.Dataset(data_loader, transform_fn)\n",
     "quantized_model = nncf.quantize(\n",
     "    model,\n",
     "    calibration_dataset,\n",
-    "    target_device=nncf.TargetDevice.CPU,\n",
     "    # Do not quantize LeakyReLU activations to allow the INT8 model to run on Intel GPU\n",
-    "    ignored_scope=nncf.IgnoredScope(patterns=[\"{re}.*LeakyReLU*\"])\n",
+    "    ignored_scope=nncf.IgnoredScope(patterns=[\".*LeakyReLU.*\"])\n",
```
eaidova, line 490: @aleksandr-mokrov these changes in ignored scope are expected or made accidentally?

aleksandr-mokrov: They are expected, from @alexsu52

eaidova (approved):

PR #1068, opened 2023-05-18
ImageBind notebook

eaidova:


adrianboguszewski:
image

eaidova:
> 0 at the end is not needed?
> 
> image

@adrianboguszewski it is new syntax introduced in 2023.0 for python API as possible replacement of compiled_model.output(0) as key. I can restore 2022.3 support if it is necessary for now

adrianboguszewski:
@eaidova, yes, please restore 2022.3 support before we release 2023.0 and update it in the requirements.

eaidova:
> @eaidova, yes, please restore 2022.3 support before we release 2023.0 and update it in the requirements.

restored

ryanloney:
I was able to run the notebook on my 10th Gen Core CPU and Arc A770 GPU

This looks a bit messy and not ideal for supporting long term. Is this really the only option for getting pre-requisites to work together? 
image

ryanloney (approved):

eaidova:
> I was able to run the notebook on my 10th Gen Core CPU and Arc A770 GPU
> 
> This looks a bit messy and not ideal for supporting long term. Is this really the only option for getting pre-requisites to work together? image

It is exact what usually hided in requirements.txt files with conditional requirements, I can hide it in requirements.txt, but not sure that it is ideal option too, because it does not allow run notebook without this file in isolated environment (e.g. google collab)

eaidova:
> I'd recommend changing the colormap not to be so random: image
> 
> Otherwise, it's fine

changes colormap, is it better now?

adrianboguszewski (approved):
It's better :)

PR #1089, opened 2023-05-29
Fixed docs conversion

adrianboguszewski:


adrianboguszewski:
@eaidova just to be aware, I'm fixing also some notebooks for documentation generation

PR #1095, opened 2023-05-31
Unignore 404-style-transfer notebook

AnishaUdayakumar:
Unignoring the 404-style-transfer notebook in CI pipeline. This change is intended to allow CI checks to run on this notebook, allowing to debug and ensure its functionality.


eaidova:
@AnishaUdayakumar it is scipped due to hangs on ci, so please make sure that it works

adrianboguszewski:
@eaidova , yes, we uncommented this to check if we can fix it somehow. We're not gonna merge it until we're sure it's fixed.

Does it hang randomly or every time?

eaidova:
> @eaidova , yes, we uncommented this to check if we can fix it somehow. We're not gonna merge it until we're sure it's fixed.
> 
> Does it hang randomly or every time?

here example of job https://github.com/openvinotoolkit/openvino_notebooks/actions/runs/5130633162/jobs/9229665657
cancelled by timeout on ubuntu20 with python3.8 the last executed notebook 404.

The most times see it is reproducible on notebooks conversion, it has the same os and python version parameters.

PR #1107, opened 2023-06-01
Adapt notebooks for devcloud

dnoliver:
NOTE: This is intended to be on a "devcloud" branch, not main!

This pull request contains 2 changes:
1. Remove pip install instructions: in DevCloud, or in environments like The Littlest JupyterHub, kernels are prepared by an admin for the users, so pip install instructions usually don't have the intended effect. Also, for a non admin users it install packages in ~/.local, which causes duplication
2. Update the kernel to be the one provided in DevCloud: this avoids having to select a kernel from a long list of available kernels, and avoid the case when somebody picks the wrong kernel and the notebook doesn't work.

These 2 changes are applied to the following notebooks:
1. 001
2. 002
3. 003
4. 004
5. 231
6. 235
7. 236
8. 240


dnoliver:
Will be handled with other method

PR #1113, opened 2023-06-05
[Gradio] Add gradio interface to 236-stable-diffusion-v2

itrushkin:
- [x] Infinite zoom notebook
- [x] Text-to-image notebook

ticket: _CVS-108281_


eaidova:
@itrushkin please preserve output in notebook, we need it for documentation building

eaidova (approved):

eaidova:
@itrushkin is it possible to stay inference without gradio before gradio launching for all notebooks where it was provided before

The reason that we lost results in notebooks documentation as gradio interface can not be embedded into docs?

PR #1114, opened 2023-06-06
Adding Google Colab support

igor-davidyuk:
This PR introduces changes to some of the notebooks to adapt them for Colab.


notebooks/utils/notebook_utils.py
```diff
@@ -13,18 +13,13 @@
 from pathlib import Path
 from typing import List, NamedTuple, Optional, Tuple
 
-import cv2
```
eaidova, line 16: opencv still used in VideoPlayer, please restore import

igor-davidyuk: The import is called inside the `VideoPlayer` function

eaidova: but ci logs speak not https://github.com/openvinotoolkit/openvino_notebooks/actions/runs/5267602177/jobs/9523076146?pr=1114

igor-davidyuk: I put it in a wrong place 😂
It should work now

eaidova (approved):

PR #21, opened 2021-04-12
Merging README to develop

raymondlo84:
@ryanloney  ... It seems the develop is behind? Maybe use this pull request to pull this forward?

PR #1118, opened 2023-06-07
update 105 with new pytorch frontend

OpenVINO-dev-contest:
1. Fix some issues in 105
2. update with new pytorch frontend


notebooks/105-language-quantize-bert/105-language-quantize-bert.ipynb
```diff
@@ -1,6 +1,7 @@
 {
```
eaidova, line 1: <pre class="ql-syntax" spellcheck="false">Line #21.        core = ov.Core()
</pre><p>we have risk to reference core before assignment below as it created only if model converted in previous run</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1118/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

eaidova, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    for input_layer in quantized_model.inputs:
</pre><p>does it still required to reshape model? probably we can convert model with preserving dynamic shapes and pass them only into benchmark app. As I can see, other part of notebook uses dynamic shapes (I checked nncf works)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1118/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

OpenVINO-dev-contest: <p>the problem is NNCF POT will not support a model with dynamic input shape, so I provide a extra parameter: "input=input_info" during model conversion. otherwise some error message will be in output.</p>

eaidova: I tested with 2.5, it supports

OpenVINO-dev-contest: <p>yes you are right, I will move it out of  if else loop.</p>

OpenVINO-dev-contest: fixed

eaidova (approved):

PR #1133, opened 2023-06-13
Removing data from the repo - part 1

igor-davidyuk:
This PR removes the `data/` folder and all of its contents. The notebooks fetch data samples from the web instead.

I follow this algorithm for this PR:
1. Identify the most extensive file in the `/notebooks/data/` folder and remove it.
2. Identify all the notebooks that rely on this data piece.

For every dependent notebook:

4. Fetch the data piece, and all other data samples in the example, from some source on the web. If none is found, fetch from [our s3](https://storage.openvinotoolkit.org/repositories/openvino_notebooks/data/data/)
5. Isolate requirements installation in the notebook.
6. Try running in Colab, fetching the `utils` package, or introducing other required modifications. _Sometimes it gets to lower video resolution to achieve fair performance in Colab, sometimes lowering the number of samples in a validation set allows keeping notebook execution time within reasonable limits._
7. If it works in Colab - update the examples README and global README adding Colab badges.

**BEFORE MERGE all the links should point to the main branch**

Ticket: CVS-112642


igor-davidyuk:
ready for review

notebooks/utils/notebook_utils.py
```diff
@@ -206,6 +208,8 @@ class VideoPlayer:
 
     def __init__(self, source, size=None, flip=False, fps=None, skip_first_frames=0):
         import cv2
+
+        self.cv2 = cv2  # This is done to access the package in class methods
```
eaidova, line 212: as we know which specific function we will use from module, can we jus import them and store as class attributes?

notebooks/utils/notebook_utils.ipynb
```diff
@@ -1,6 +1,7 @@
 {
```
eaidova, line 1: <p>please restore outputs for utility notebook, it requires for understanding utils content without reading code</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1133/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

notebooks/utils/notebook_utils.py
igor-davidyuk: why would we do this? The whole package will be loaded anyway + we need more fields in the class

eaidova: possibly prevent illegal usage of package inside class, when user uses field instead of direct module usage. I do not insist, just try to reduce number of questions to the code design.

BTW, what happens if uses does not install opencv at all, should we provide some warning for this case or import error will be enough clear from UX perspective?

igor-davidyuk: These are good questions. In my understanding, currently, we manage package installation directly within the example notebooks. It is the easiest approach, but I believe that the introduction of an import-checking mechanism of some kind would reduce the clarity and transparency of the code, and would be somewhat redundant than helpful.

notebooks/utils/notebook_utils.ipynb
igor-davidyuk: I restored the outputs!

notebooks/205-vision-background-removal/205-vision-background-removal.ipynb
```diff
@@ -57,12 +71,36 @@
...
+    "if not utils_file_path.exists():\n",
+    "    !git clone --depth 1 https://github.com/openvinotoolkit/openvino_notebooks.git\n",
+    "    utils_file_path = Path(\"./openvino_notebooks/notebooks/utils/notebook_utils.py\")\n",
+    "    notebook_directory_path = Path(\"./openvino_notebooks/notebooks/205-vision-background-removal/\")\n",
+    "\n",
+    "sys.path.append(str(utils_file_path.parent))\n",
+    "sys.path.append(str(notebook_directory_path))\n",
+    "\n",
+    "from notebook_utils import load_image\n",
+    "from model.u2net import U2NET, U2NETP"
```
apaniukov, line 99: Why clone the whole repository? The `notebook_utils.py` and `u2net.py` can be downloaded with urllib, like in other notebooks.

notebooks/405-paddle-ocr-webcam/405-paddle-ocr-webcam.ipynb
```diff
@@ -64,18 +57,38 @@
...
+   "source": [
+    "# Import local modules\n",
+    "\n",
+    "utils_file_path = Path('../utils/notebook_utils.py')\n",
+    "notebook_directory_path = Path('.')\n",
+    "\n",
+    "if not utils_file_path.exists():\n",
+    "    !git clone --depth 1 https://github.com/igor-davidyuk/openvino_notebooks.git -b moving_data_to_cloud openvino_notebooks\n",
+    "    utils_file_path = Path('./openvino_notebooks/notebooks/utils/notebook_utils.py')\n",
+    "    notebook_directory_path = Path('./openvino_notebooks/notebooks/405-paddle-ocr-webcam/')\n",
```
apaniukov, line 81: Same here.

apaniukov (approved):

notebooks/205-vision-background-removal/205-vision-background-removal.ipynb
apaniukov: 205 and 405 notebooks rely on one file each. Only 407 is the problem. We could leave this as it is for now, but fix it next time.

igor-davidyuk: In my opinion, cloning the repo is a cleaner and more reliable approach than retrieving a file with a URL request. I only did this for the utils module to save users' data bandwidth

PR #1159, opened 2023-06-29
Throughput performance tricks notebooks

adrianboguszewski:
The second notebook with the throughput tricks

PR #1164, opened 2023-07-02
[Gradio] Add Gradio interface to 214-grammar-correction

itrushkin:
This PR introduces interactive inference for Grammar correction notebook using Gradio.

ticket: _CVS-108281_


notebooks/214-grammar-correction/214-grammar-correction.ipynb
```diff
@@ -62,12 +62,12 @@
   },
```
igor-davidyuk, line 62: <p>!pip install -q 'openvino-dev&gt;=2023.0.0'</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1164/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

itrushkin: <p><code>openvino</code>,<code>onnx</code> and <code>onnxruntime</code> packages are installed with <code>optimum-intel[openvino]</code> . So this string becomes <code>!pip install "optimum-intel[openvino] @ git+https://github.com/huggingface/optimum-intel.git" gradio</code>  <span class="ql-cursor">﻿</span></p>

igor-davidyuk:
For some reason, the last cell does not work for me
Screenshot 2023-07-02 151832

notebooks/214-grammar-correction/214-grammar-correction.ipynb
eaidova: @itrushkin  @igor-davidyuk  it was made intentionally, optimum pin specific version of openvino and its update schedule is not aligned with us, so it may lead to misalignment and accidental reinstalling openvino during optimum install (with upgrading or downgrading version), that can be unexpected, so installing separately is better choise

eaidova:
> For some reason, the last cell does not work for me Screenshot 2023-07-02 151832

@igor-davidyuk please check proxy settings

itrushkin:
@eaidova Can you confirm it works on your side?

notebooks/214-grammar-correction/214-grammar-correction.ipynb
itrushkin: Added installation of OpenVINO pinned to `2023.0.0`

eaidova:
> @eaidova Can you confirm it works on your side?

on my end it works
image

Regarding the issue reported by @igor-davidyuk, I sow this some time ago and in my case, it was connected to the usage intel proxy setting in environment without VPN (similar issue on gradio side https://github.com/gradio-app/gradio/issues/3330)

eaidova:
BTW, the flag button is used for what? It does not looks like something changed when I click it and there is no any description how it can be used

itrushkin:
> BTW, the flag button is used for what? It does not looks like something changed when I click it and there is no any description how it can be used

@eaidova From Gradio docs:
>When a user testing your model sees input with interesting output, they can click the flag button to send the input and output data back to the machine where the demo is running. The sample is saved to a CSV log file (by default). If the demo involves images, audio, video, or other types of files, these are saved separately in a parallel directory and the paths to these files are saved in the CSV file.

More: https://gradio.app/using-flagging/

I don't think we should disable this, should we?

eaidova:
> > BTW, the flag button is used for what? It does not looks like something changed when I click it and there is no any description how it can be used
> 
> @eaidova From Gradio docs:
> 
> > When a user testing your model sees input with interesting output, they can click the flag button to send the input and output data back to the machine where the demo is running. The sample is saved to a CSV log file (by default). If the demo involves images, audio, video, or other types of files, these are saved separately in a parallel directory and the paths to these files are saved in the CSV file.
> 
> More: https://gradio.app/using-flagging/
> 
> I don't think we should disable this, should we?

I do not know, I think we need to add some description for that at least (for now, from user perspective, who do not familiar to gradio, it is not clear how it should be used, but not sure that it is very useful in our case...

notebooks/214-grammar-correction/214-grammar-correction.ipynb
eaidova, line 62: <pre class="ql-syntax" spellcheck="false">Line #1.    import ipywidgets as widgets
</pre><p>I think as user has opportunity to modify text in gradio demo, there is no need in widget, just use plain text instead, otherwise we have 2 different ways to launch model with user input</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1164/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

itrushkin: Will it affect the project documentation?

eaidova: what do you mean under documentation?
Probably it should not, it just removing dependency of notebook from pywidgets.

In part before running radio you just make sure that pipeline work, so constant input should be ok for these purposes as you have gradio for providing input. If under documentation, you mean description inside notebook, probably it requires small changes with removing sentence which says that user should provide own text in widget window (I do not remember exactly)

itrushkin: Done

eaidova (approved):

adrianboguszewski:
@eaidova, Windows CI doesn't work
image

eaidova:
> @eaidova, Windows CI doesn't work image

@adrianboguszewski  @itrushkin Probably it is because running gradio in debug=True, other platforms hanging and aborted by timeout, can we turn it to false during test patching?

adrianboguszewski:
@eaidova I noticed it doesn't start in the #1159 also and I don't use gradio there. I think it's not gradio-related, unfortunately :/

aleksandr-mokrov (approved):

notebooks/214-grammar-correction/214-grammar-correction.ipynb
```diff
@@ -89,7 +89,7 @@
     }
```
eaidova, line 89: <pre class="ql-syntax" spellcheck="false">Line #1.    !pip install -q "git+https://github.com/huggingface/optimum-intel.git" openvino==2023.0.0 onnx onnxruntime gradio
</pre><p>&nbsp;"openvino&gt;=2023.0.0 "</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1164/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

itrushkin: Done

PR #1166, opened 2023-07-03
update ov on prerelease 2023.1.0

eaidova:


eaidova:
failed bug for 216 notebook against tf lite fe - CVS-114332

PR #1174, opened 2023-07-06
fix: output of postprocess function, converting to numpy array

NguyenJoe0201:
The output returned by the postprocess function is inconsistent. As far as I know, the output should be a numpy array, so I added the `numpy()` method to convert it to numpy array.


notebooks/230-yolov8-optimization/230-yolov8-optimization.ipynb
```diff
@@ -584,7 +584,7 @@
     "            continue\n",
     "        if proto is None:\n",
     "            pred[:, :4] = ops.scale_boxes(input_hw, pred[:, :4], shape).round()\n",
-    "            results.append({\"det\": pred})\n",
+    "            results.append({\"det\": pred.numpy()})\n",
```
NguyenJoe0201, line 587: return a numpy array

PR #1180, opened 2023-07-11
Small fixes in README

adrianboguszewski:


eaidova:
@adrianboguszewski  I do not think that your changes correct. our requirements txt does not support python3.11 and some notebooks too (there is no paddlepaddle package for python3.11 as well), please return back python3.10 as higher version.

adrianboguszewski:
@eaidova fixed in #1183

PR #1163, opened 2023-06-30
Add named entity recognition notebook

sbalandi:
[CVS-113297](https://jira.devtools.intel.com/browse/CVS-113297)

TODO: change link on colab to develop branch main repo before merge

PR #1194, opened 2023-07-17
move ner notebook to main

eaidova:


eaidova:
moved from https://github.com/openvinotoolkit/openvino_notebooks/pull/1163

PR #1243, opened 2023-08-14
Ignore 249-oneformer-segmentation

sgolebiewski-intel:
Excluding 244-named-entity-recognition.ipynb from conversion.


andrei-kochin (changes requested):
Please explain what is the reason to exclude

adrianboguszewski:
@andrei-kochin This is excluded only for one stage of documentation conversion because it causes timeouts. It means this notebook won't be re-run to generate new outputs and the current outputs will be used. It doesn't impact anything else (CI, etc.)

andrei-kochin (approved):

adrianboguszewski (approved):

PR #1269, opened 2023-08-25
Adding notebook on POT sparsity and quantization

ronandorrepaal:


notebooks/122-pytorch-post_training_quantization_and_sparsity_pot/122-pytorch-post-training-quantization-sparsity-pot-public.ipynb
```diff
@@ -0,0 +1,668 @@
+{
```
eaidova, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    # On Windows, this script adds the directory that contains cl.exe to the PATH to enable PyTorch to find the
</pre><p>does it really required? as I know this part of code in previous notebooks used for NNCF quantization using PyTorch to be able load custom pytorch layers realization, it is not used in your notebook.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1269/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

eaidova, line 1: <p>does notebook should somehow mention that prepared model can be run on NPU?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1269/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

MaximProshin:
I wonder why we spend time on the notebook with POT if it's deprecated since 2023.0 and will most likely be removed in 2024.0? Moreover the sparsity support in POT was always experimental but fully supported in NNCF. 
So I highly recommend to replace POT by NNCF here for both PTQ and Sparsity.
One more thing for Sparsity is to get the benefit of it in OpenVINO, one needs to enable it in the runtime by CPU_SPARSE_WEIGHTS_DECOMPRESSION_RATE. See an example here: https://github.com/openvinotoolkit/openvino_notebooks/blob/main/notebooks/116-sparsity-optimization/116-sparsity-optimization.ipynb

MaximProshin (changes requested):
To merge this notebook, the following issues should be sorted out:
1. POT is deprecated and will be removed. All OV compression notebooks should rely on NNCF.
2. POT Sparsity is not exposed to users and not supported formally (originally due to absence of HW support, proper validation, etc). OV marketing was always against of documenting it in OV. Only NNCF supports it formally.
3. I don't understand how OV users should get any benefit from Sparsity in this notebook.

MaximProshin:
The model you take in this notebook is actually taken from another notebook with NNCF: 
fp32_pth_url = "https://storage.openvinotoolkit.org/repositories/nncf/openvino_notebook_ckpts/304_resnet50_fp32.pth"

NNCF already has such model quantized and sparsed with different levels here: https://github.com/openvinotoolkit/nncf/blob/develop/docs/ModelZoo.md#pytorch-classification
For example, 
• QAT: INT8
• Sparsity: 50% (RB)

It means the notebook can be quickly switched to NNCF.

PR #1292, opened 2023-09-06
Detectron2 conversion tutorial

eaidova:
CVS-119678


eaidova:
<p>as we recommend to run notebooks in own environment, there is no need in cuda. Cuda specific version requires extra 800mb from user storage (and in some pytorch version +2Gb for installing cuda libs), so by default we recommend using CPU version </p>

---

 <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1292/discussion/#comment-1712753642'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

eaidova:
<p>actually it is OVDict (but it is not trivial how to import it from package, so that is why I use its parent class). OVDict supports tensor names, indexes and openvino output tensors as keys.</p>

---

 <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1292/discussion/#comment-1712805518'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

igor-davidyuk:
> as we recommend to run notebooks in own environment, there is no need in cuda. Cuda specific version requires extra 800mb from user storage (and in some pytorch version +2Gb for installing cuda libs), so by default we recommend using CPU version
> 
> [View entire conversation](https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1292/discussion/#comment-1712753642) on ReviewNB

It feels strange to gatekeep users from installing the most popular DL framework in the world, which they probably already have installed and duplicated in numerous virtual environments 😁
Plus, avoiding these `sys_platform` based constructions would streamline our code so much.

eaidova:
<p>if you want, you can chenge pytorch version in the next release in all notebooks, for now, I prefer to keep it aligned for all cases.</p>

Issue #1301, opened 2023-09-08
236-stable-diffussion-v2-text-to-image-demo can't run


dnoliver:
**Describe the bug**
A clear and concise description of what the bug is, including steps to reproduce the bug if possible.

236-stable-diffussion-v2-text-to-image-demo can't run

In this cell https://github.com/openvinotoolkit/openvino_notebooks/blob/main/notebooks/236-stable-diffusion-v2/236-stable-diffusion-v2-text-to-image-demo.ipynb?short_path=427ad71#L196-L209, the values for the paths return empty, and then the models can't be loaded in further cells.

In this script https://github.com/openvinotoolkit/openvino_notebooks/blob/main/notebooks/236-stable-diffusion-v2/implementation/ov_stable_diffusion_pipeline.py#L12, there is a bad import of openvino, Results in a crash

 **Expected behavior**

Should work end to end

**Screenshots**

First problem:

![image](https://github.com/openvinotoolkit/openvino_notebooks/assets/7730267/7b7a8fe0-7398-4872-8ad7-9ce973e1b37c)

works after defining the paths

```python
txt_encoder_ov_path="./sd2.1/text_encoder.xml"
unet_ov_path="./sd2.1/unet.xml"
vae_encoder_ov_path="./sd2.1/vae_encoder.xml"
vae_decoder_ov_path="./sd2.1/vae_decoder.xml"
```

Second problem:

![image](https://github.com/openvinotoolkit/openvino_notebooks/assets/7730267/fbd5bb0f-cecc-451e-9661-891c20d8a940)

Works after changing the import to `import openvino as ov`

 **Installation instructions (Please mark the checkbox)**
[x] I followed the installation guide at https://github.com/openvinotoolkit/openvino_notebooks#-installation-guide to install the notebooks. 

** Environment information **
Please run `python check_install.py` in the _openvino_notebooks_ directory. If the output is _NOT OK_ for any of the checks, please follow the instructions to fix that. If that does not work, or if you still encounter the issue, please paste the output of check_install.py here.

```bash
root@ubuntu:/opt/intel/openvino_notebooks# python3 check_install.py  
Traceback (most recent call last):                                                                                                                                        
  File "/opt/intel/openvino_notebooks/check_install.py", line 7, in <module> 
    import pip                                                                       
ModuleNotFoundError: No module named 'pip'                                           
[...]

PR #1308, opened 2023-09-12
fix sd 2.1 demo

eaidova:


eaidova:
fix for https://github.com/openvinotoolkit/openvino_notebooks/issues/1301

PR #1352, opened 2023-09-27
use fp32 data type for vae on gpu in SD 1.5

sshlyapn:


sshlyapn:
@eaidova , could you please take a look?

eaidova (approved):

PR #1361, opened 2023-10-04
Fix dataset downloading for YOLOv8 notebook

vshampor:
The config files for datasets in ultralytics (such as `coco.yaml`) contain the executable code for downloading the dataset, which gets invoked, in particular, in the `check_det_dataset` function. This code is dependent on the installed ultralytics version, while the version of the `coco.yaml` downloaded in the notebook was not pinned. 

Furthermore, the `check_det_dataset` internally checks for the presence of the dataset at a specific directory (`DATASETS_DIR`), which is in general different from the `OUT_DIR` and as a result the dataset would be downloaded twice. Fixed that too.


vshampor:
@eaidova please advise whether we really need the direct download/extraction of the dataset with the `OUT_DIR` and related code, or whether we can do this entirely via the `check_det_dataset` which seems to download the dataset as well if it is missing in the specified data folder

eaidova:
> @eaidova please advise whether we really need the direct download/extraction of the dataset with the `OUT_DIR` and related code, or whether we can do this entirely via the `check_det_dataset` which seems to download the dataset as well if it is missing in the specified data folder

@vshampor  thank you for the fix! 
on the one hand, reusing existing ultralytics API for dataset downloading can be beneficial, but from the other point of view explicit dataset downloading allows users to easily replace dataset with their own one

eaidova (approved):

PR #1375, opened 2023-10-11
Fixing memory issues 236 Optimun-Intel demos

paularamo:
- Release memory in two notebooks: 
1- https://github.com/paularamo/openvino_notebooks/blob/main/notebooks/236-stable-diffusion-v2/236-stable-diffusion-v2-optimum-demo-comparison.ipynb
2- https://github.com/paularamo/openvino_notebooks/blob/main/notebooks/236-stable-diffusion-v2/236-stable-diffusion-v2-optimum-demo.ipynb

- Improve the description of the notebook in the 236-readme file. 
3- https://github.com/paularamo/openvino_notebooks/blob/main/notebooks/236-stable-diffusion-v2/README.md


eaidova:
@paularamo thanks, what do you think about adding the same disclamer about running once and that it may requires GPU into demo-comparision inside notebook?

Notebooks integrated into openvino docs as independent pages without README, in case if user just see this notebook in docs, it maybe useful to have info about notebook goal inside

eaidova (approved):

Issue #1373, opened 2023-10-10
Intelligent Queue Manager Recipe - utils.py gets the incorrect file size if the file is compressed


dnoliver:
**Describe the bug**
A clear and concise description of what the bug is, including steps to reproduce the bug if possible.

In this line https://github.com/openvinotoolkit/openvino_notebooks/blob/recipes/recipes/intelligent_queue_management/utils.py#L156-L158 (in the `recipes` branch), when the code do the request, it gets the file size in the 'Content-length' header.

The value is then used in this line https://github.com/openvinotoolkit/openvino_notebooks/blob/recipes/recipes/intelligent_queue_management/utils.py#L171.

When the file is compressed in the remote server, the size reported by the first call is the size of the compressed file, not the actual file size. This triggers the re-download of the file all the time.

In this line, the sample gets a file from raw.githubusercontent.com https://github.com/openvinotoolkit/openvino_notebooks/blob/recipes/recipes/intelligent_queue_management/convert_and_optimize.py#L18, which happens to be compressed, so the `coco.yaml` file always get downloaded

This behavior is described in https://github.com/psf/requests/issues/5319#issuecomment-575946480

It can be avoided by setting the Accept-Encoding: Identity flag in the first call:

```
response = requests.get(
            url=url, headers={"User-agent": "Mozilla/5.0", "Accept-Encoding": "identity"}, stream=True
        )
```

 **Expected behavior**

Files should not be downloaded if they are already in the target location

**Screenshots**

In the following text (kind of ugly), you can see how coco.yaml gets the wrong size and triggers the download

```
[...]

PR #1376, opened 2023-10-11
Update utils.py

AnishaUdayakumar:
Modified the `requests.get` function in `download_file` to include the "Accept-Encoding": "identity" header. This change ensures that the 'Content-length' header will return the actual file size, not the compressed size. This fixes the issue where files would be re-downloaded due to a mismatch in reported and actual file sizes.
Please refer: https://github.com/openvinotoolkit/openvino_notebooks/issues/1373


AnishaUdayakumar (commented):
Removed the commented lines

adrianboguszewski (approved):

PR #1389, opened 2023-10-18
Fix keeping CI artifacts

smirnov-alexey:
Option wasn't working before due to type mismatch. Fixed the option, works as expected now


smirnov-alexey:
@eaidova could you merge it please?

eaidova (approved):

PR #1411, opened 2023-10-31
Added quantization to LCM notebook

l-bat:
Add post-training optimization support by NNCF for UNet model from [263-latent-consistency-models-image-generation.ipynb](https://github.com/openvinotoolkit/openvino_notebooks/blob/main/notebooks/263-latent-consistency-models-image-generation/263-latent-consistency-models-image-generation.ipynb)
Pipeline performance speed up: 1.307

PR #1413, opened 2023-11-01
FastSAM: add quantization

igor-davidyuk:
This PR introduces the quantization step to 261.
CVS-123720


MaximProshin:
@nikita-savelyevv , please also take a look

sbalandi (approved):

notebooks/261-fast-segment-anything/261-fast-segment-anything.ipynb
```diff
@@ -18,6 +18,8 @@
     "- [Original Pipeline Inference](#FastSAM-in-Ultralytics)\n",
```
nikita-savelyevv, line 18: <blockquote>moSaveSerialize OpenVINO IR model, usov.save_modelserialize` function. </blockquote><p>Something's wrong here.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1413/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

nikita-savelyevv, line 18: <pre class="ql-syntax" spellcheck="false">Line #1.    %pip install -q nncf
</pre><p>Maybe move to the top?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1413/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

nikita-savelyevv, line 18: <p>Could you please add a reference to this example <a href="https://github.com/openvinotoolkit/nncf/tree/develop/examples/post_training_quantization/openvino/yolov8" target="_blank">https://github.com/openvinotoolkit/nncf/tree/develop/examples/post_training_quantization/openvino/yolov8</a> since the quantization code was taken from there?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1413/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

nikita-savelyevv, line 18: <p>Usually, the quantization is made optional by using (1) a widget to enable it and (2) <code>skip</code> extension from <code>utils/skip_kernel_extension.py</code> to control cell execution based on the widget value. Please refer for an example to PR #1411 .</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1413/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

nikita-savelyevv, line 18: <p>Benchmark app does not always represent the performance boost accurately. For example in this case, FPS increased 4 times, however if you would run the model here, the performance boost would be less than 2x. </p><p> </p><p>Because of this I think it would be better to compare performance by running the real pipeline here, for example on calibration dataset.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1413/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

nikita-savelyevv, line 18: <p>As I understand, after your changes the interactive demo will still use the original model and not the quantized one. </p><p> </p><p>Usually, when quantization is added, the interactive demo is also changed to visually demonstrate the comparison between the models. You can refer to PR #1411 for an example of such demo changes.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1413/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

igor-davidyuk: <p>fixed!</p>

igor-davidyuk: <p>Unfortunately, this installation can not be skipped with the help of skip_kernel_extension.py. So I moved the installation to the top of the notebook</p>

igor-davidyuk: <p>added a mention </p>

igor-davidyuk: <p>integrated the extention, thank you!</p>

igor-davidyuk: <p>Done!</p>

igor-davidyuk: <p>I added a radio button allowing to switch between the quantized and regular models</p>

eaidova, line 18: <pre class="ql-syntax" spellcheck="false">Line #3.    %pip install -q nncf
</pre><p>"nncf&gt;2.6.0"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1413/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

eaidova (approved):

notebooks/261-fast-segment-anything/261-fast-segment-anything.ipynb
igor-davidyuk: done!

nikita-savelyevv: <p>Thanks! Could you please also add an additional output for the time it took to make prediction? Otherwise benefits of quantization are not reflected in the demo.</p>

nikita-savelyevv: <p>Also, seems like radio button needs to be hidden in case quantization is not enabled. Otherwise fp32 model is executed nevertheless of radio button value.</p>

igor-davidyuk: > Also, seems like radio button needs to be hidden in case quantization is not enabled. Otherwise fp32 model is executed nevertheless of radio button value.

I see it as over-complication. What if the user has the flag unchecked but the quantized checkpoint is present? And what if the flag is checked but the checkpoint is not there? I think it is good enough right now. @eaidova what do you think?

nikita-savelyevv: > > Thanks! Could you please also add an additional output for the time it took to make prediction? Otherwise benefits of quantization are not reflected in the demo.
> 
> Gradio indicates the latest inference time

Yes, I've seen it in some cases. But here I don't see it:
image

MaximProshin:
Please check if it works on iGPU. Recent several notebooks with int8 don't work there. For this one I see empty bounding boxes on iGPU even for FP32 model.

notebooks/261-fast-segment-anything/261-fast-segment-anything.ipynb
aleksandr-mokrov: In fact I don't see any difference in time execution in gradio (about 1-2 seconds). Looks like gradio brings a lot of overhead costs and we can only demonstrate that we have the same output for both models.

nikita-savelyevv: I see. Then indeed no need to add a time output.

nikita-savelyevv: > > Also, seems like radio button needs to be hidden in case quantization is not enabled. Otherwise fp32 model is executed nevertheless of radio button value.
> 
> I see it as over-complication. What if the user has the flag unchecked but the quantized checkpoint is present? And what if the flag is checked but the checkpoint is not there? I think it is good enough right now. @eaidova what do you think?

These are the valid questions.

> What if the user has the flag unchecked but the quantized checkpoint is present?

Then we don't load the quantized model and show it in demo.

> And what if the flag is checked but the checkpoint is not there?

This should error out, as it currently will. Because `quantized_wrapped_model` either will not be declared or would not be able to be loaded.

> I think it is good enough right now. 

Sorry, I disagree. When I select quantized model, I expect the quantized model to be executed, not the original one.

I suggest the following change:

```
model_type = gr.Radio(
            ["FP32 model", "Quantized model"] if do_quantize.value else ["FP32 model"],
            value="FP32 model", label="Select model variant"
        )
```
I don't think it makes it much more complicated compared to how it is now.

nikita-savelyevv, line 18: <pre class="ql-syntax" spellcheck="false">Line #12.    for image in tqdm(coco_dataset):
</pre><p>Maybe add description to <code>tqdm</code> like "Measuring inference time"?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1413/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

nikita-savelyevv, line 18: <pre class="ql-syntax" spellcheck="false">Line #7.    for image in tqdm(coco_dataset):
</pre><p>Maybe add description to <code>tqdm</code> like "Measuring inference time"?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1413/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

aleksandr-mokrov (approved):

notebooks/261-fast-segment-anything/261-fast-segment-anything.ipynb
igor-davidyuk: added!

igor-davidyuk:
> Please check if it works on iGPU. Recent several notebooks with int8 don't work there. For this one I see empty bounding boxes on iGPU even for FP32 model.

I confirm, the model returns no result on Windows + iGPU

notebooks/261-fast-segment-anything/261-fast-segment-anything.ipynb
igor-davidyuk: added!

igor-davidyuk: Added proposed code block

nikita-savelyevv (approved):
Thanks!

MaximProshin:
> > Please check if it works on iGPU. Recent several notebooks with int8 don't work there. For this one I see empty bounding boxes on iGPU even for FP32 model.
> 
> I confirm, the model returns no result on Windows + iGPU

@igor-davidyuk , was it addressed somehow, by a bug or changes in the notebook?

eaidova:
> > > Please check if it works on iGPU. Recent several notebooks with int8 don't work there. For this one I see empty bounding boxes on iGPU even for FP32 model.
> > 
> > 
> > I confirm, the model returns no result on Windows + iGPU
> 
> @igor-davidyuk , was it addressed somehow, by a bug or changes in the notebook?

@MaximProshin we have bug for original SAM model inference on GPU

PR #1419, opened 2023-11-03
Fix LCM demo

l-bat:
Interactive demo crashed if `Randomize seed across runs` is False. Because the order of `inputs` arguments in `gr.on` is inconsistent with `generate` arguments.
image


l-bat:
@eaidova please take a look

eaidova (approved):

PR #1423, opened 2023-11-05
distil whisper notebook

eaidova:
CVS-124239


yury-gorbachev (approved):
This is very good notebook, I have sent my version with very minor updates separately to accelerate process a bit since I cannot rerun in my env.

PR #1428, opened 2023-11-06
Added int4 compression to LLM chatbot notebook

nikita-savelyevv:
Added int4 compression to **llama-2-chat-7b** and **mpt-7b-chat** model in LLM chatbot notebook. Performance boost is about 1.16x for both (compared to INT8).

Also, disabled INT4 and FP16 conversion for **red-pajama-3b-chat** because it currently fails due to 123973.

Ticket 124179


OpenVINO-dev-contest:
Hi, I just wonder: compared with GPTQ, which compression method is our first recommendation to openvino users

notebooks/254-llm-chatbot/254-llm-chatbot.ipynb
```diff
@@ -55,10 +55,10 @@
    "source": [
```
eaidova, line 55: <pre class="ql-syntax" spellcheck="false">Line #5.        value=model_ids[0],
</pre><p>could you please change index by default on llama or mpt as we have some issues with red-pijama (that is not solved yet) and this sample is to so great for demonstrating int4</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1428/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

eaidova, line 55: <p>need to add note that dGPU currently does not support INT4</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1428/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

eaidova, line 55: <pre class="ql-syntax" spellcheck="false">Line #10.    compression_mode
</pre><p>can we convert several converted precision in once or only one always should can be selected?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1428/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

MaximProshin: <p>I think it should be stated differently, like "There is no speedup for INT4 compressed models on dGPU". </p>

MaximProshin, line 55: <p>Compress model weights to INT8 precision using&nbsp;<a href="https://github.com/openvinotoolkit/nncf" target="_blank">NNCF</a> =&gt; Compress model weights to INT8 <strong>or INT4</strong> precision using&nbsp;<a href="https://github.com/openvinotoolkit/nncf" target="_blank">NNCF</a></p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1428/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

MaximProshin, line 55: <p><code>nncf.compress_weights</code> function accept =&gt; <code>nncf.compress_weights</code> function accept<strong>s</strong> OpenVINO model instance and compress<strong>es</strong></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1428/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

nikita-savelyevv: Yes, it would just be an extra work considering that only one of the model will be used in the end. 

Do you suggest to replace dropdown menu with multiple checkboxes?

eaidova: I'm not sure, does it makes sense? From comparison point of view it will be convenient to compress models at once, but if later we need to select anyway which model we want to use, I think it maybe more stright-forward to use your proposed style with only single option without extra complicated logic

nikita-savelyevv: <p>Done</p>

nikita-savelyevv: <p>Made MPT the default option because running llama is quire complicated requiring to gain access rights.</p>

nikita-savelyevv: <p>Done</p>

nikita-savelyevv: <p>Done</p>

nikita-savelyevv: <p>Sometime it makes sense and doesn't bring any complications for uses that don't need it. I'll add it. :+1:</p>

eaidova (approved):

notebooks/254-llm-chatbot/254-llm-chatbot.ipynb
AlexKoff88: <p>again, I would use the notion of "8-bit or 4-bit data types" to make it more generic.</p>

AlexKoff88: <p>The same story for INT8 on dGPU.</p>

AlexKoff88:
> Hi, Im just wondering: compared with GPTQ, which compression method is our first recommendation to openvino users

We recommend 8-bit compression by default because it does not drop accuracy in general cases. What kind of INT4 to use depends on the model's size. You can find some examples [here](https://github.com/openvinotoolkit/nncf/blob/22e7d982701e4ddd6e3429831bd0d465e2f089cb/docs/compression_algorithms/CompressWeights.md).

notebooks/254-llm-chatbot/254-llm-chatbot.ipynb
nikita-savelyevv: <p>Changed</p>

notebooks/254-llm-chatbot/254-llm-chatbot.ipynb
```diff
@@ -479,38 +617,6 @@
...
-    }
-   ],
-   "source": [
-    "int8_compressed_weights = widgets.Checkbox(\n",
-    "    value=True,\n",
-    "    description='Use compressed weights',\n",
-    "    disabled=False\n",
-    ")\n",
-    "\n",
-    "widgets.VBox([device, int8_compressed_weights])"
```
aleksandr-mokrov, line 511: Add `device` row. After removing this block, there no device output widget.

notebooks/254-llm-chatbot/254-llm-chatbot.ipynb
```diff
@@ -710,7 +815,13 @@
     "from optimum.intel.openvino import OVModelForCausalLM\n",
     "from transformers import AutoTokenizer\n",
     "\n",
-    "model_dir = Path(model_id.value) / (\"FP16\" if not int8_compressed_weights.value else \"INT8_compressed_weights\")\n",
+    "if prepare_int4_model.value:\n",
```
aleksandr-mokrov, line 818: On the previous stage it is possible to choose all compression model types. But here we get the first chosen. Looks like it is make sense to add widget for choosing what type of converted models we want to use.

notebooks/254-llm-chatbot/254-llm-chatbot.ipynb
eaidova: INT8 for dGPU gives speedup but not significantly as I know, but ok

eaidova: Possibly it is better to rephrase there is no speedup -> there may not be speedup

nikita-savelyevv: Rephrased

notebooks/254-llm-chatbot/254-llm-chatbot.ipynb
nikita-savelyevv: Thanks, fixed

notebooks/254-llm-chatbot/254-llm-chatbot.ipynb
nikita-savelyevv: Added a dropdown for selecting a model

aleksandr-mokrov (approved):

MaximProshin (approved):
LGTM. checked int4 mpt on CPU and iGPU. I noticed that it's slower on iGPU but I believe it's not a notebook problem

PR #1430, opened 2023-11-07
Fix LCM notebook due to diffusers API change

ngaloppo:
The LCM scheduler was refactored sometime between diffusers 0.21.4 an 0.22.1, which broke the LCM notebook. Fixing the notebook to 0.21.4 until the notebook gets rewritten to the new scheduler API.


ngaloppo:
Closing since LCM isn't even supported until diffusers 0.22.0. I'm not sure how this worked before, but it's broken for me now. We'll need to find an alternative fix.

PR #1431, opened 2023-11-07
Fix LCM notebook after upstream model update

ngaloppo:
The upstream model for LCM_Dreamshaper_v7 got updated to work with the LCM pipeline from diffusers 0.22.0. This broke the notebook.

As suggested by the [model card](https://huggingface.co/SimianLuo/LCM_Dreamshaper_v7#usage-deprecated) README, it is required to check out a specific revision of the model to continue using the old community pipelines. This is applied in this commit, until the notebook gets upgraded to the 0.22.0 diffusers approach.


eaidova (approved):

eaidova:
@ngaloppo thank you, I'm working on updating on official diffusers pipeline right now

PR #1447, opened 2023-11-09
add zephyr support in chatbot notebook

eaidova:


CalebXDonoho:
Hello, in running the notebook on HuggingFaceH4/zephyr-7b-beta, I noticed that fp16 and int8 convert; however, int4 runs into an nncf error as shown attached- claiming that the conversion to int4 cannot happen due to lack of support on bf16.I have been running these tests on SPR HBM Max 9480 if that helps! Are there additional setup steps I should be taking or is the int4 quantization not possible on this machine? Thanks for your help! nncf_conversion_error

ryanloney:
I was able to run on my 10th Gen Core Windows laptop with 32GB of RAM and generate int4 compressed IR, then run on CPU. Trying iGPU next. 
image

MaximProshin:
> Hello, in running the notebook on HuggingFaceH4/zephyr-7b-beta, I noticed that fp16 and int8 convert; however, int4 runs into an nncf error as shown attached- claiming that the conversion to int4 cannot happen due to lack of support on bf16.I have been running these tests on SPR HBM Max 9480 if that helps! Are there additional setup steps I should be taking or is the int4 quantization not possible on this machine? Thanks for your help! nncf_conversion_error

I guess it's SPR specifics where bf16 was enabled in the conversion step. @eaidova , is it possible to avoid it?

eaidova:
@CalebXDonoho  could you please try to replace installation optimum-intel to my branch instead of provided in notebook ()?

%pip install git+https://github.com/eaidova/optimum-intel.git@ea/fp32_dtype

if it helps, then I'll submit fix to optimum-intel

brmarkus:
I have seen (early and exotic variants of with early BSP, BIOS) SPR CPUs "not supporting BF16"; `lscpu` hasn't revealed the corresponding instruction sets...

eaidova:
@CalebXDonoho could you please try to execute following code in your env and share with me results:

```python
import torch
print(torch.get_default_dtype())
```
Looks like on your system the defult dtype set to torch.bfloat16 that lead to converting model with preserving this dtype

CalebXDonoho:
hello @eaidova . This is what I see when attempting the steps discussed prior on my SPR HBM upon reinstall optimum-intel using the steps provided. This should also show the default dtype I am using. Let me know if I need to do further tests
screenshot_1110

PR #1463, opened 2023-11-16
move chatglm patch out of converter.py

OpenVINO-dev-contest:
move chatglm patch out of ```converter.py```, otherwise chatglm conversion will fail


OpenVINO-dev-contest:
@eaidova please take a look

eaidova (approved):

PR #1485, opened 2023-11-22
Set seed in LCM notebook for quantization

l-bat:
The seed is necessary for stable reproduction of quantization results


eaidova (approved):

l-bat:
diffusers 0.23.1 contains some fixes for `LCMScheduler`. According to my experiments with quantization, we can achieve better results with the latest version.

PR #1500, opened 2023-11-30
mixed precision infer for RedPajama

pavel-esir:
TBD

For some reason still there are errors while calibrating on GPU
```
While validating node 'opset1::Concat Concat_884256 () -> ()' with friendly_name 'Concat_884256':
Shape inference input shapes {[1,32,0,80],[0,32,15,80]}
Argument shapes are inconsistent; they must have the same rank, and must have equal dimension everywhere except on the concatenation axis (axis 2).
```


pavel-esir:
closing since already implemented in https://github.com/openvinotoolkit/openvino_notebooks/pull/1602

PR #1508, opened 2023-12-01
Added quantization to SDXL-Turbo notebook

l-bat:
image


eaidova:
@l-bat  sdxl is quit heavy network in comparison with SD, maybe it make sense to separated demo run (I mean not only side by side running 2 pipelines mode, but also select only 1 fp or quantized model) before demo?

PR #1520, opened 2023-12-07
Added support for a Japanese LLM model, rinna/youri-7b-chat

yas-sim:
I added support for a Japanese LLM model, rinna/youri-7b-chat. The base architecture of the model is Llama2.

Note: The response from the LLM model is a bit shorter than I expected. It may be because of the strict system prompt message. If I use a more relaxed and short system prompt, the model response will become longer.


eaidova:
@yas-sim thank you for your contribution! It is great to see Japanese LLM support in chatbot notebook!

Could you please provide short description for model in README and notebook?

yas-sim:
Ok. I have added a short description of Rinna-7b-chat model to README.md and the notebook. Please check them.

PR #1525, opened 2023-12-08
Added quantization to Wuerstchen notebook

l-bat:


MaximProshin:
@l-bat , in all recent notebooks we talk about disabling "FastBiasCorrection" but any BiasCorrection is disabled actually, so let's rephrase it. Also we explain this optional param but not others. E.g. in this notebook we also use model_type, so let's explain it too.  As for the ignored scope, explain why these layers were ignored or at least say that "according to our experiments these layers impact the accuracy dramatically".

notebooks/265-wuerstchen-image-generation/265-wuerstchen-image-generation.ipynb
```diff
@@ -780,7 +780,7 @@
...
     "\n",
-    "According to `WuerstchenPriorPipeline` structure, prior model is used in the cycle repeating inference on each diffusion step, while text encoder takes part only once. According to `WuerstchenDecoderPipeline` structure, decoder model is used in the cycle repeating inference on each diffusion step, while other parts of pipeline take part only once. That is why computation cost and speed of prior and decoder models become the critical path in the pipeline. Quantizing the rest of the pipeline does not significantly improve inference performance but can lead to a substantial degradation of accuracy.\n",
+    "According to `WuerstchenPriorPipeline` structure, prior model is used in the cycle repeating inference on each diffusion step, while text encoder takes part only once, and in the `WuerstchenDecoderPipeline`, the decoder model is used in a loop, and other pipeline components infer only once. That is why computation cost and speed of prior and decoder models become the critical path in the pipeline. Quantizing the rest of the pipeline does not significantly improve inference performance but can lead to a substantial degradation of accuracy.\n",
```
nikita-savelyevv, line 783:
```suggestion
    "According to `WuerstchenPriorPipeline` structure, prior model is used in the cycle repeating inference on each diffusion step, while text encoder takes part only once, and in the `WuerstchenDecoderPipeline`, the decoder model is used in a loop, and other pipeline components are inferred only once. That is why computation cost and speed of prior and decoder models become the critical path in the pipeline. Quantizing the rest of the pipeline does not significantly improve inference performance but can lead to a substantial degradation of accuracy.\n",
```

nikita-savelyevv (changes requested):

nikita-savelyevv (approved):

MaximProshin (approved):

PR #1526, opened 2023-12-08
[DOCS] Adding refurbished images for notebooks in the docs

sgolebiewski-intel:
Adding refurbished images (with a 1:1 aspect ratio) that will be used as thumbnails for each notebook presented in the documentation.


kblaszczak-intel:
There are still some minor corrections on the way and all images will be reduced in size to 150x150

eaidova:
@sgolebiewski-intel, thanks preview looks good, except one - 209. it is difficult to understand what shown on it may be better to have left corner of image instead of center crop and also more space around it (having smaller font size for recognized text, now it is difficult to see where image and where text that was recognized on it)

Also organizational question. Who should take care about new added notebooks preview images? If contributors, please add note about this requirements in [CONTRIBUTING.md](https://github.com/openvinotoolkit/openvino_notebooks/blob/main/CONTRIBUTING.md). We are also prefer do not store images in the repo and use github for them, but as it is small size and as I can see do not used in readme files, I think it is ok to stay them here

sgolebiewski-intel:
> @sgolebiewski-intel, thanks preview looks good, except one - 209. it is difficult to understand what shown on it may be better to have left corner of image instead of center crop and also more space around it (having smaller font size for recognized text, now it is difficult to see where image and where text that was recognized on it)
> 
> Also organizational question. Who should take care about new added notebooks preview images? If contributors, please add note about this requirements in [CONTRIBUTING.md](https://github.com/openvinotoolkit/openvino_notebooks/blob/main/CONTRIBUTING.md). We are also prefer to not store images in the repo and use github for them, but as it is small size and as I can see do not used in readme files, I think it is ok to stay them here

I believe you meant 208, right? I have produced an update for that picture, with a top left part of it being cropped, as per your suggestion.

eaidova:
no, I mean https://github.com/openvinotoolkit/openvino_notebooks/blob/ef4f7cee32db0b3e0e08803b8fdad3a88b216b57/notebooks/209-handwritten-ocr/209-handwritten-ocr.png

sgolebiewski-intel:
> no, I mean https://github.com/openvinotoolkit/openvino_notebooks/blob/ef4f7cee32db0b3e0e08803b8fdad3a88b216b57/notebooks/209-handwritten-ocr/209-handwritten-ocr.png

In this picture, I cropped the right part, which with the glyphs included is more meaningful than the initial part. From what I have ascertained it means "Let's just do it" so I thought of it as a good example for a thumbnail. 
With your suggestion, I would propose:
image
I have resized the source text and slanted it a bit, to create a more deliberate sense of a written passage. What do you think?

kblaszczak-intel:
When it goes to maintaining the images, we can deal with cropping and resizing, if the initial images are provided. About storing them in the notebook repository, as soon as we change one thing in the documentation build CI, we'll be able to move the images to the main repo.

eaidova:
>> I have resized the source text and slanted it a bit, to create a more deliberate sense of a written passage. What do you think?

yes, thanks, it looks better

eaidova (approved):

PR #1527, opened 2023-12-11
Add device target to ssd 1b lcm compile_model

epsilonError:
Adds selected device target to `compile_model` method calls for LCM distilled version of segmind/SSD-1B.

Addition to #1519.


eaidova:
@epsilonError thanks!

PR #1543, opened 2023-12-14
[254-rag-chatbot]Separate the inference devices for embedding and LLM model

OpenVINO-dev-contest:
1. separate the device selection for embedd and LLM model
2. update the stop tokens of chatglm3
3. fix the stop tokens missing issue


eaidova (approved):

andrei-kochin:
Fixes #1550

PR #1593, opened 2024-01-04
[204] Unmatching types - correction of timm library version

Roszczyk:
Notebook didn't work, because of changes in timm library. Segmenter library, which is used as well, hasn't been updated since 2022. New versions of timm library use now different definition of DefaultCfg class. 

To fix unmatching types without updating Segmenter library, we have to use timm version 0.4.12.

[Fixing a bug: CVS-127425]


adrianboguszewski (approved):

eaidova (approved):

eaidova:
@Roszczyk thank you for the fix, please also add notebook into https://github.com/openvinotoolkit/openvino_notebooks/blob/main/.ci/ignore_pip_conflicts.txt
as now it required to clone repo before checking deps, unfortunately, such behaviour does not supported by pip conflict checker yet

Roszczyk:
@eaidova Done. 204 added to ignore pip conflicts in the last commit.

PR #1609, opened 2024-01-11
Update ignored node names

aleksandr-mokrov:
CVS-127423
Ignored node names were changed for pre-release.


notebooks/107-speech-recognition-quantization/107-speech-recognition-quantization-data2vec.ipynb
```diff
@@ -69,7 +69,8 @@
    "outputs": [],
    "source": [
     "%pip uninstall -q -y openvino-dev openvino openvino-nightly\n",
-    "%pip install -q \"openvino-nightly\" \"nncf>=2.7\"\n",
+    "%pip install -q --pre openvino\n",
```
eaidova, line 72: there is no prerelease 2023.3 on pypi, so this change will install 2023.2 instead and notebook fails as ignored scope different.
https://pypi.org/project/openvino/#history
it is better to stay nightly for now

PR #1615, opened 2024-01-12
Updates for speechT5 TTS Model

AnishaUdayakumar:
Inclusion of load_tts.py and update of app.py for SpeechT5 TTS model


recipes/conversational_voice_agent/load_tts.py
```diff
@@ -0,0 +1,26 @@
+import nltk
```
adrianboguszewski, line 1: This file seems not to be needed, as exactly the same is done in app.py. 

However, when you add openvino conversion, it should be done in a separate file.

recipes/conversational_voice_agent/requirements.txt
```diff
@@ -14,4 +14,9 @@ librosa==0.10.1
 gradio==4.3.0
 
 # Bark model from the GitHub repository
-git+https://github.com/suno-ai/bark.git@773624d26db84278a55aacae9a16d7b25fbccab8
\ No newline at end of file
+git+https://github.com/suno-ai/bark.git@773624d26db84278a55aacae9a16d7b25fbccab8
```
adrianboguszewski, line 17: Bark requirement and bark_utils.py are no longer needed.

recipes/conversational_voice_agent/requirements.txt
```diff
@@ -14,4 +14,9 @@ librosa==0.10.1
...
+
+# SpeechT5
+datasets
+torch
+soundfile
```
adrianboguszewski, line 22: All dependencies must be pinned in this kind of project. We don't want to wake up and realize something doesn't work due to updates.

recipes/conversational_voice_agent/app.py
```diff
@@ -203,14 +239,14 @@ def create_UI(initial_message: str) -> gr.Blocks:
...
-            # submit button
+	    # submit button
             submit_audio_btn = gr.Button("Submit", variant="primary", scale=1)
-
-        # chatbot
+	
+	# chatbot
         chatbot_ui = gr.Chatbot(value=[[None, initial_message]], label="Chatbot")
-        # chatbot's audio response
+	# chatbot's audio response
```
adrianboguszewski, line 249: Something is wrong with these indentations. They shouldn't change

recipes/conversational_voice_agent/app.py
```diff
@@ -146,16 +154,44 @@ def chat(history: List) -> List[List[str]]:
...
+            speaker_embeddings = torch.tensor(embeddings_dataset[7306]["xvector"]).unsqueeze(0)
+            speech = tts_model.generate_speech(inputs["input_ids"], speaker_embeddings, vocoder=tts_vocoder)
+            audio_segments.append(speech.numpy())
+    
+    # Combine audio segments
+    combined_audio = np.concatenate(audio_segments, axis=0) if audio_segments else np.array([])
+    end_time = time.time()  # End time
+    print("TTS model synthesis time: {:.2f} seconds".format(end_time - start_time))
+    
+    return 16000, combined_audio
```
adrianboguszewski, line 194: Please return AUDIO_WIDGET_SAMPLE_RATE instead of a magic number.

recipes/conversational_voice_agent/app.py
```diff
@@ -146,16 +154,44 @@ def chat(history: List) -> List[List[str]]:
...
-    # generate audio for the last prompt (bot's response)
-    return SAMPLE_RATE, tts_model.generate_audio(prompt)
+
+    # Function to split text into sentences
+    def split_into_sentences(text):
+        sentences = re.split(r'(?<=[^A-Z].[.!?]) +(?=[A-Z])', text)
+        return [sentence.strip() for sentence in sentences if sentence.strip()]
+
+    # Split the prompt into sentences
+    sentences = split_into_sentences(prompt)
```
adrianboguszewski, line 176: It's better to use here nltk tokenization (look at bark_utils file) than using own function.

recipes/conversational_voice_agent/app.py
```diff
@@ -146,16 +154,44 @@ def chat(history: List) -> List[List[str]]:
...
+    
+    # Combine audio segments
+    combined_audio = np.concatenate(audio_segments, axis=0) if audio_segments else np.array([])
+    end_time = time.time()  # End time
+    print("TTS model synthesis time: {:.2f} seconds".format(end_time - start_time))
+    
+    return 16000, combined_audio
 
 
 def transcribe(audio: Tuple[int, np.ndarray], conversation: List[List[str]]) -> List[List[str]]:
```
adrianboguszewski, line 197: Please add time measurement for ASR also. It's a nice feature!

adrianboguszewski (changes requested):

recipes/conversational_voice_agent/requirements.txt
AnishaUdayakumar: @adrianboguszewski you mean we need to specify the version for each dependency or remove the versions?

AnishaUdayakumar (commented):
I have some issue while  pushing the commits. Kindly request to review the changes or advise if I should create a new PR.

PR #1626, opened 2024-01-18
fix: output images aren't displayed for stable diffusion

DimaPastushenkov:
Move removal of the pipelines and the garbage collector to the separate cells. Otherwise output images aren't displayed.


eaidova (approved):

eaidova:
@DimaPastushenkov thanks, @paularamo just fyi

PR #1630, opened 2024-01-19
Stable diffusion with IP-Adapter notebook

eaidova:
TO DO

- [x] readme
- [x] table of content
- [x] text


notebooks/278-stable-diffusion-ip-adapter/278-stable-diffusion-ip-adapter.ipynb
aleksandr-mokrov, line 1: Probably it makes sense to install accelerate?:
```
Cannot initialize model with low cpu memory usage because `accelerate` was not found in the environment. Defaulting to `low_cpu_mem_usage=False`. It is strongly recommended to install `accelerate` for faster and less memory-intense model loading. You can do so with: 

pip install accelerate
```

aleksandr-mokrov, line 1: ModuleNotFoundError: No module named 'cv2'

aleksandr-mokrov, line 1: Can you add example in gradio?

eaidova: it does not make sense :) low cpu memory for initialization used only in pair with offloading model on cuda, as we stay model on cpu during conversion and need full weights initialization, it is useless and only introduce additional dependencies

eaidova: added to dependencies install

eaidova: added

PR #1665, opened 2024-01-30
Updated with native speechT5 Model changes

AnishaUdayakumar:
Updated with native speechT5 Model changes including app.py and requirements.txt, removed bark_utils.py


recipes/conversational_voice_agent/app.py
```diff
@@ -146,16 +153,44 @@ def chat(history: List) -> List[List[str]]:
...
     """
+    start_time = time.time()  # Start time
     prompt = conversation[-1][-1]
-    # generate audio for the last prompt (bot's response)
-    return SAMPLE_RATE, tts_model.generate_audio(prompt)
+
+    # Function to split text into sentences
+    def split_into_sentences(text):
+        sentences = re.split(r'(?<=[^A-Z].[.!?]) +(?=[A-Z])', text)
+        return [sentence.strip() for sentence in sentences if sentence.strip()]
```
adrianboguszewski, line 172: Use nltk package here

recipes/conversational_voice_agent/app.py
```diff
@@ -146,16 +153,44 @@ def chat(history: List) -> List[List[str]]:
...
+
+    # Split the prompt into sentences
+    sentences = split_into_sentences(prompt)
+    audio_segments = []
+
+    for sentence in sentences:
+        inputs = tts_processor(text=sentence, return_tensors="pt")
+
+        # Check if the token count is within the limit
+        if inputs.input_ids.size(1) <= 600:
```
adrianboguszewski, line 182: What happens if the input is longer?

recipes/conversational_voice_agent/app.py
```diff
@@ -146,16 +153,44 @@ def chat(history: List) -> List[List[str]]:
...
+            embeddings_dataset = load_dataset("Matthijs/cmu-arctic-xvectors", split="validation")
+            speaker_embeddings = torch.tensor(embeddings_dataset[7306]["xvector"]).unsqueeze(0)
```
adrianboguszewski, line 184: This is very inefficient. You should load it once, not in every loop.

recipes/conversational_voice_agent/app.py
```diff
@@ -146,16 +153,44 @@ def chat(history: List) -> List[List[str]]:
...
+            speech = tts_model.generate_speech(inputs["input_ids"], speaker_embeddings, vocoder=tts_vocoder)
+            audio_segments.append(speech.numpy())
+
+    # Combine audio segments
+    combined_audio = np.concatenate(audio_segments, axis=0) if audio_segments else np.array([])
```
adrianboguszewski, line 189: This "if" seems to be unnecessary. `audio_segments` always exists.

recipes/conversational_voice_agent/load_tts.py
```diff
@@ -0,0 +1,26 @@
+import nltk
+from nltk.tokenize import word_tokenize
+
+# Attempt to download the 'punkt' tokenizer
+try:
+    nltk.data.find('tokenizers/punkt')
+except LookupError:
+    nltk.download('punkt')
```
adrianboguszewski, line 8: Put it into separate function and call it from main

adrianboguszewski (changes requested):
Good job @AnishaUdayakumar. It works good. Please see the comments below

recipes/conversational_voice_agent/app.py
AnishaUdayakumar: I will address by implementing the chunk_sentence function. This function breaks down sentences that exceed the token limit into smaller parts. Each chunk is then separately processed by the TTS model.

recipes/conversational_voice_agent/app.py
AnishaUdayakumar: The revised synthesize function directly concatenates audio_segments without the conditional check.

recipes/conversational_voice_agent/load_tts.py
AnishaUdayakumar: Moved the NLTK to app.py

PR #1668, opened 2024-01-31
Reorganizing table of contents for better conversion to docs

sgolebiewski-intel:
The aim of the PR is to separate the "Table of Contents" from blocks of text, to avoid issues of incorrect formatting, being the result of conversion of  *.ipynb to *.rst files. To resolve the problems with formatting in the converted files, I would suggest the following changes in the source files:

* introducing blank lines, that will precede and follow the `#### Table of contents:\n` string, 
* or keeping its contents in a single cell, when the initial cell includes both an extensive introductory paragraph and the aforementioned table of contents). 

The file changed in the [016cd73ca07f1c1a07fe0d70b5a931328fe81790](https://github.com/openvinotoolkit/openvino_notebooks/pull/1668/commits/016cd73ca07f1c1a07fe0d70b5a931328fe81790) commit is an example of the recurrent problem after converting the notebooks.  It will serve as a POC for now. @eaidova, @sbalandi, please, share your thoughts.


adrianboguszewski (approved):
LGTM

andrei-kochin:
@sgolebiewski-intel have to revert in #1694 as generation script which generates the table of contents was not updated

Please discuss with @eaidova

PR #1682, opened 2024-02-02
[406] Using notebook_utils functions instead of OMZ

Roszczyk:
So far done:

1. notebook_utils downloaded
2. model downloaded via download_file function from notebook_utils

to do:

1. pth to OpenVINO IR conversion without OMZ


Roszczyk:
those unused libraries that make the code check fail - will be used (hopefully) to prepare conversion from pth to onnx and to OpenVINO IR

PR #1716, opened 2024-02-15
integrate minicpm into chatbot notebook

eaidova:


notebooks/254-llm-chatbot/254-rag-chatbot.ipynb
```diff
@@ -392,34 +391,29 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from converter import converters\n",
```
OpenVINO-dev-contest, line 395: why converters is removed ?

eaidova: I also noted that and fixed in another commit, it is just wrong copy-paste

OpenVINO-dev-contest:
besides the import, there are something deleted may cause errors:
image

PR #1727, opened 2024-02-19
Update LLM notebooks to transition to weight compression via optimum-intel

nikita-savelyevv:


notebooks/275-llm-question-answering/275-llm-question-answering.ipynb
```diff
@@ -182,7 +182,7 @@
     "### Weights Compression using Optimum Intel\n",
```
AlexKoff88, line 182: <pre class="ql-syntax" spellcheck="false">Line #41.            "mistral-7b": {
</pre><p>Recently, we introduced such <a href="https://github.com/huggingface/optimum-intel/blob/8f7d016fbeb0cd185e80b9852207c2b530d3d394/optimum/intel/openvino/configuration.py#L81" target="_blank">configs</a> in Optimum-Intel so they should work OOB. Can you please check if we can drop them from the notebooks?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1727/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

nikita-savelyevv: Seems like the configs you reference serve only as a recommendation and in fact the provided `quantization_config` is used for compression and not the `default_config` : https://github.com/huggingface/optimum-intel/blob/8f7d016fbeb0cd185e80b9852207c2b530d3d394/optimum/intel/openvino/modeling_decoder.py#L611

AlexKoff88 (approved):

eaidova (approved):

PR #1745, opened 2024-02-23
update langchain's api

OpenVINO-dev-contest:
As ```chain.run()``` will be removed in langchain 0.2.0, it needs be replaced with ```chain.invoke()```

### DIFF
--- a/notebooks/254-llm-chatbot/254-rag-chatbot.ipynb
+++ b/notebooks/254-llm-chatbot/254-rag-chatbot.ipynb
@@ -1346,7 +1346,7 @@
     "    stream_complete = Event()\n",
     "\n",
     "    def infer(question):\n",
-    "        rag_chain.run(question)\n",
+    "        rag_chain.invoke(question)\n",
     "        stream_complete.set()\n",
     "\n",
     "    t1 = Thread(target=infer, args=(history[-1][0],))\n",


notebooks/254-llm-chatbot/254-llm-chatbot.ipynb
```diff
@@ -510,7 +510,7 @@
     "            \"group_size\": 128,\n",
     "            \"ratio\": 0.8,\n",
     "        },\n",
-    "        \"chatglm2-6b\": {\n",
+    "        \"chatglm3-6b\": {\n",
```
OpenVINO-dev-contest, line 513: hi @eaidova could you also help to confirm chatglm3's quantization configurations ?

eaidova: it should be the same like for changpt2, @AlexKoff88  @ljaljushkin is it correct?

eaidova (approved):

notebooks/254-llm-chatbot/254-llm-chatbot.ipynb
ljaljushkin: @eaidova we didn't search the optimal quantization config for chatglm3. Probably, it should be as chatglm2, since the default ratio r=1.0 might be too high.

eaidova: chatglm3 is fine-tuned with more data version of chatglm2, that is why I say that here should not be changes

PR #1789, opened 2024-03-05
230-YOLOv8 (Validate Quantized model Inference "GPU addition" was not working) Fixed

paularamo:
I have found a bug with the latest update managing ov_config. I'm fixing this as soon as possible since this notebooks has been using for external and internal folks. Please take a look at the error message in the comments. Thanks and Best Regards


eaidova:
sorry, but your fix is not correct

PR #1809, opened 2024-03-08
add filter for non-rag model

OpenVINO-dev-contest:
Remove redundant codes in qwen and chatglm wrapper.


OpenVINO-dev-contest:
@eaidova could you help to check, thanks

eaidova (approved):

PR #1857, opened 2024-03-28
Use legacy keras & update seed

as-suvorov:


eaidova:
@as-suvorov  could you please also update image in notebook meta?
it looks like now it uses image from clip notebook...

image

as-suvorov:
> @as-suvorov could you please also update image in notebook meta? it looks like now it uses image from clip notebook...
> 
> image

I need a kind of checklist for adding new notebooks, so many things to miss:)
Thank you! fixed

eaidova (approved):

PR #1891, opened 2024-04-05
Add NNCF optimization to stable-diffusion-keras-cv notebook

l-bat:


MaximProshin (approved):
@l-bat , maybe we shouldn't say "Hybrid" in the list of steps to avoid any confusion and just say Quantization. Up to you.

eaidova:
@l-bat  looks like after adding quantization, there is not enough RAM for finishing notebook on github runners, could you please add it into ignored list?

eaidova (approved):

nikita-savelyevv (approved):

PR #133, opened 2021-06-29
Create README for 301-tensorflow-training-openvino

ryanloney:
Added a README for the two notebooks in 301-tensorflow

PR #1901, opened 2024-04-09
migrate llm chatbot notebook on optimum conversion approach

eaidova:


notebooks/llm-chatbot/llm-chatbot.ipynb
```diff
@@ -1,7 +1,6 @@
 {
```
OpenVINO-dev-contest, line 1: <pre class="ql-syntax" spellcheck="false">Line #98.            "qwen-7b-chat": {"sym": True, "group_size": 128, "ratio": 0.6},
</pre><p>it should be reformatted to align with other config</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1901/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

OpenVINO-dev-contest, line 1: <pre class="ql-syntax" spellcheck="false">Line #133.        ov_model.save_pretrained(int4_model_dir)
</pre><p>Can we save tokenizer to local as well, so we can run the Gradio demo without network.</p><p> </p><p>BTW,  I wonder if we can use <code>cli</code> tool to export the remote model ? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/1901/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

eaidova: cli tool can be used for export models with remote code, it required to add --trust_remote_code as cmd parameter. Due to a lot of parameters variativity (like trust_remote_code or int4 quantization config with sym/asym switching) I think using cli in notebook can become overcomplicated and hard for understanding how to select parameters for a user, that is why I think using model class is the better choice here, but I'll add note that cli is also can be used for export.

OpenVINO-dev-contest: I believe our users do not wish to spend to much efforts on model exporting, so `cli` can be a "one-stop" tool for them.

PR #1935, opened 2024-04-17
Align optimum quantization calls with the latest API

nikita-savelyevv:
Recently, there were some [changes ](https://github.com/huggingface/optimum-intel/pull/638)made to how NNCF quantization is called via optimum-intel. This PR updates the quantization use cases to how they are currently intended to be called.


eaidova (approved):

eaidova:
@nikita-savelyevv  thanks! I wanted to ask about this update today  :)

PR #1944, opened 2024-04-18
Quantized sparse resnet

ronandorrepaal:
Adding training aware sparse quantization notebook based on imagenet. Data set must be downloaded separately, as per requirements


mlyashko:
@eaidova , could you please take a look?

eaidova:
@ronandorrepaal looks like there is some rebasing issues in your branch (I have suspicious that you started work with main branch before refactoring and then creates pr to latest), it is difficult to review due to a lo of irrelevant changes reported. Could you please resubmit pr using latest branch as base for your changes?

PR #1973, opened 2024-04-26
Update README.md

catchygit:
update custom AI assistant gif


adrianboguszewski (changes requested):
Could we remove these black strips from the GIF? It would look much better :)

adrianboguszewski (approved):

PR #2006, opened 2024-05-08
Update patch_notebooks.py to include 'openvino'

wkobielx:
I believe we are missing removing `openvino` from notebooks while patching, what leads  to uninstalling openvino during tests:

example on `test_llm-agent-langchain.ipynb` after patching:
image

This leads to next tests failing due to missing `openvino` package.

### DIFF
--- a/.ci/patch_notebooks.py
+++ b/.ci/patch_notebooks.py
@@ -61,7 +61,7 @@ def remove_ov_install(cell):
                     continue
                 if "openvino-tokenizers" in part:
                     continue
-                if "openvino>" in part or "openvino=" in part:
+                if "openvino>" in part or "openvino=" in part or "openvino" in part:
                     continue
                 if empty:
                     empty = not has_additional_deps(part)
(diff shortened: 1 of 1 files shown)


wkobielx:
After fix:
image

wkobielx:
@eaidova can you take a look?

aleksandr-mokrov:
@wkobielx, this fix also affects nncf (git+https://github.com/openvinotoolkit/nncf.git). Looks like it should be a strict comparison `"openvino" == part`

wkobielx:
@aleksandr-mokrov updated

wkobielx:
image
image
image

few examples, seems it is not affecting nncf repo anymore

wkobielx:
@aleksandr-mokrov @yatarkan can you take a look? There is a high chance that a lot of tests will fail with missing `openvino` package on next nightly run without this fix

yatarkan (approved):

PR #2010, opened 2024-05-08
add filter for unsupported devices

OpenVINO-dev-contest:
It will help to remove NPU option when running LLM.


notebooks/llava-multimodal-chatbot/llava-multimodal-chatbot.ipynb
```diff
@@ -1187,9 +1187,12 @@
     "\n",
```
eaidova, line 1187: <pre class="ql-syntax" spellcheck="false">Line #5.    support_devices = ["CPU", "GPU"]
</pre><p>that will not work in systems with multiple GPUs (in case of several devices available GPU, they will be named as GPU.0, GPU.1 and so on)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/2010/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

eaidova: IMHO, better to filter != NPU instead

eaidova (approved):

PR #2021, opened 2024-05-16
update the GPU searching function in multi GPUs

openvino-book:
The original code works well in single GPU mode, and will work wrong in Multi GPUs mode, so I update the code from
```
if "GPU" not in core.available_devices:
```
to
```
if any('GPU' in device for device in core.available_devices) == False:
```

### DIFF
--- a/notebooks/auto-device/auto-device.ipynb
+++ b/notebooks/auto-device/auto-device.ipynb
@@ -87,7 +87,7 @@
     "\n",
     "core = ov.Core()\n",
     "\n",
-    "if any('GPU' in device for device in core.available_devices) == False:\n",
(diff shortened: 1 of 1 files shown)


eaidova (approved):

notebooks/auto-device/auto-device.ipynb
```diff
@@ -87,7 +87,7 @@
     "\n",
     "core = ov.Core()\n",
     "\n",
-    "if \"GPU\" not in core.available_devices:\n",
+    "if any('GPU' in device for device in core.available_devices) == False:\n",
```
eaidova, line 90:
```suggestion
    "if not any('GPU' in device for device in core.available_devices):\n",
```

PR #2066, opened 2024-06-04
add Qwen2 sample

OpenVINO-dev-contest:


OpenVINO-dev-contest:
@eaidova since we are preparing for a marketing PR before qwen2 officially released , could you help to review this Pull Request today. Thanks

supplementary_materials/qwen2/requirements.txt
```diff
@@ -0,0 +1,7 @@
+nncf==2.10.0
+optimum-intel>=1.16.0
+transformers>=4.38.2
+openvino>=2024.1.0
```
eaidova, line 4: can we use nightly for obtaining the latest optimizations?

supplementary_materials/qwen2/requirements.txt
```diff
@@ -0,0 +1,7 @@
+nncf==2.10.0
+optimum-intel>=1.16.0
```
eaidova, line 2: 1.16.0 does not contain config for qwen2 if I right remember
```suggestion
optimum-intel>=1.17.0
```

supplementary_materials/qwen2/requirements.txt
OpenVINO-dev-contest: no, as my test, the latest nightly build has accuracy issue on qwen2

supplementary_materials/qwen2/convert.py
```diff
@@ -0,0 +1,74 @@
...
+    elif args.precision == "int8":
+        ov_model = OVModelForCausalLM.from_pretrained(model_path, export=True,
+                                                      compile=False, load_in_8bit=True)
+    else:
+        ov_model = OVModelForCausalLM.from_pretrained(model_path, export=True,
+                                                      compile=False, load_in_8bit=False)
+
+    ov_model.save_pretrained(ir_model_path)
+
+    print("====Exporting tokenizer=====")
```
eaidova, line 71: this message may lead to confusion (especially, because we have ov tokenizers that are not used here), because you do not export it, just same on disk.

BTW, possibly it is better to include ov tokenizer to be able run model with https://github.com/openvinotoolkit/openvino.genai/tree/master/text_generation/causal_lm/cpp

supplementary_materials/qwen2/convert.py
```diff
@@ -0,0 +1,74 @@
...
+    tokenizer = AutoTokenizer.from_pretrained(
+        model_path, trust_remote_code=True)
```
eaidova, line 73: there is no need in key `trust_remote_code` of this sample cover only qwen2, model does not have remote code

supplementary_materials/qwen2/convert.py
```diff
@@ -0,0 +1,74 @@
...
+    parser.add_argument('-p',
+                        '--precision',
+                        required=False,
+                        default="int4",
+                        type=str,
+                        choices=["fp16", "int8", "int4"],
+                        help='fp16, int8 or int4')
+    parser.add_argument('-o',
+                        '--output',
+                        default='./Qwen1.5-0.5B-ov',
```
eaidova, line 30: can we do not hardcode output dir and change it base on model_id if not provided? (it may lead to some confusion if they are not aligned)

eaidova:
@OpenVINO-dev-contest, generally looks good. I have some general questions here;

1. can we put notebook in the same folder to automate preparation steps described in readme?
2. can we add some UI (gradio/streamlit) version of demo?

supplementary_materials/qwen2/chat.py
```diff
@@ -0,0 +1,124 @@
+import argparse
+from typing import List, Tuple
+from threading import Thread
+import torch
```
eaidova, line 4: there is no torch listed in requirements, possibly better to add explicitly for avoid getting model without SDPA due to too old pytorch

OpenVINO-dev-contest:
> @OpenVINO-dev-contest, generally looks good. I have some general questions here;
> 
> 1. can we put notebook in the same folder to automate preparation steps described in readme?
> 2. can we add some UI (gradio/streamlit) version of demo?

Hi @OpenVINO-dev-contest I think we can do these in our chatbot notebook, and this example code is just a quick start guide, so I do not expect too many dependencies with it. It should be clean and easy to use.

supplementary_materials/qwen2/chat.py
eaidova: yes, but we did not control version here I recommend to add with lowerbound >= 2.1

eaidova (approved):

PR #2073, opened 2024-06-05
Remove device selection as only CPU can run this due to strings as input

aleksandr-mokrov:
CVS-143329


eaidova:
@aleksandr-mokrov please add notebook into this list https://github.com/openvinotoolkit/openvino_notebooks/blob/latest/.ci/check_notebooks.py#L9

eaidova (approved):

PR #2078, opened 2024-06-05
fix subset for testing in qrcode monster

eaidova:


notebooks/qrcode-monster/qrcode-monster.ipynb
```diff
@@ -1156,7 +1160,7 @@
    "execution_count": 40,
    "metadata": {
     "test_replace": {
-     "subset_size = 200": "subset_size = 4"
+     "subset_size = 200": "subset_size = 8",
```
eaidova, line 1163:
```suggestion
     "subset_size = 200": "subset_size = 8"
```

PR #2183, opened 2024-07-10
fix streamer in instruction following notebook

eaidova:


notebooks/llm-question-answering/llm-question-answering.ipynb
```diff
@@ -697,6 +698,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "import numpy as np\n",
```
eaidova:
```suggestion
```

PR #2202, opened 2024-07-16
check setuptools version upd

eaidova:


sahilgraceflight26-maker:
ea/check_packaging last year view

PR #2203, opened 2024-07-16
phi-3 vision notebook

eaidova:


notebooks/phi-3-vision/ov_phi3_vision.py
```diff
@@ -0,0 +1,541 @@
...
+
+
+def convert_phi3_model(model_id, output_dir, quantization_config):
+    output_dir = Path(output_dir)
+
+    lang_model_path = output_dir / "language_model.xml"
+    image_embed_path = output_dir / "image_embed.xml"
+    img_projection_path = output_dir / "img_projection.xml"
+    embed_token_path = output_dir / "embed_token.xml"
+
```
aleksandr-mokrov, line 201: I would suggest to add some prints for different stages that takes a long time, like "Downloading models...", "Conversion models to OpenVINO IR format...", "Quantization".

utils/notebook_utils.py
```diff
@@ -28,6 +28,27 @@
...
+
+    support_devices = core.available_devices+ ["AUTO"]
+    exclude = exclude or []
+    if exclude:
+        for ex_device in exclude:
+            if ex_device in support_devices:
+                support_devices.remove(ex_device)
+
+    device = widgets.Dropdown(
+        options=supported_devices,
```
aleksandr-mokrov, line 45: `support_devices `was defined, not `supported_devices`

eaidova: fixed

notebooks/phi-3-vision/ov_phi3_vision.py
```diff
@@ -0,0 +1,556 @@
...
+import gc
+import openvino as ov
+import torch
+from transformers import AutoModelForCausalLM
+from transformers import AutoProcessor
+from transformers.generation import GenerationConfig, GenerationMixin
+from transformers.modeling_outputs import CausalLMOutputWithPast, BaseModelOutputWithPast
+from transformers import AutoConfig
+
+from typing import Optional, Tuple, List
```
aleksandr-mokrov: There already is `typing` several lines above. And it's better to put imports in order (standard libraries, related third party imports and local )

notebooks/phi-3-vision/ov_phi3_vision.py
```diff
@@ -0,0 +1,556 @@
...
+    processor.save_pretrained(output_dir)
+    print("✅ Original model successfully loaded")
+
+    if not embed_token_path.exists():
+        print("⌛ Convert Input embedding model")
+        ov_model = ov.convert_model(model.model.embed_tokens, example_input=torch.ones([2, 2], dtype=torch.int64))
+        ov.save_model(ov_model, embed_token_path)
+        del ov_model
+        cleanup_torchscript_cache()
+        gc.collect()
```
aleksandr-mokrov, line 229: Probably it's better to move `gc.collect()` in  `cleanup_torchscript_cache`?

aleksandr-mokrov (approved):

notebooks/phi-3-vision/ov_phi3_vision.py
eaidova: explicit having command is better, because there is no guarantee that inside function it can control the same context level

PR #2216, opened 2024-07-23
Add quantization for SD3 model

l-bat:


eaidova:
@l-bat, looks like defined ignored scope is not suitable for proxy model used in tests, could you please fix notebook testing:

```
 ValidationError: Ignored nodes with name ['__module.model.base_model.model.pos_embed.proj.base_layer/aten::_convolution/Convolution'] were not found in the NNCFGraph. Refer to the original_graph.dot to discover the operationsin the model currently visible to NNCF and specify the ignored/target scopes in terms of the names there.
```

model that used on ci: https://huggingface.co/katuni4ka/tiny-random-sd3

eaidova (approved):

PR #2264, opened 2024-08-05
Fix Distl Whisper error: out of bounds for dimension 0 with size y that occurs …

aleksandr-mokrov:
CVS-148233

Problem is in pythorch 2.4.0 on windows on CPU. torch version is limited.

### DIFF
--- a/notebooks/distil-whisper-asr/distil-whisper-asr.ipynb
+++ b/notebooks/distil-whisper-asr/distil-whisper-asr.ipynb
@@ -70,7 +70,7 @@
    },
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


eaidova:
@aleksandr-mokrov could you please also apply fix for whisper notebook too? I have suspicious that issue should reproducible with this model too

aleksandr-mokrov:
> @aleksandr-mokrov could you please also apply fix for whisper notebook too? I have suspicious that issue should reproducible with this model too

You already did it a couple weeks ago:)  https://github.com/openvinotoolkit/openvino_notebooks/pull/2228/files

eaidova (approved):

PR #2278, opened 2024-08-09
Parler tts compatibility due to last updates

aleksandr-mokrov:
Support compatibility with last updates in Parler TTS without support of static cache.


eaidova (approved):

eaidova:
@aleksandr-mokrov please fix formatting for your modified notebook

PR #2285, opened 2024-08-13
add flux.1

eaidova:


notebooks/flux.1-image-generation/flux.1-image-generation.ipynb
```diff
@@ -0,0 +1,892 @@
+{
```
sbalandi, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    model_dir = convert_flux(model_selector.value)
</pre><p>it fails for me on step of Transformer conversion, with error:</p><p> </p>

<pre class="ql-syntax" spellcheck="false">OpConversionFailure: Check 'is_conversion_successful' failed at src/frontends/pytorch/src/frontend.cpp:171:
FrontEnd API failed with OpConversionFailure:
Model wasn't fully converted. Failed operations detailed log:
-- prim::Constant with a message:
String constant cannot be converted to OpenVINO opset and should be removed by consuming operation.
-- prim::ListConstruct with a message:
Exception happened during conversion of operation __module.pos_embed/prim::ListConstruct with schema (no schema)
Check '(c_node)' failed at src/frontends/pytorch/src/op/list_construct.cpp:25:
FrontEnd API failed with OpConversionFailure:
[PyTorch Frontend] Translation for prim::ListConstruct support only constant inputs

Summary:
-- normalize step failed with: Check 'input_type_0.compatible(input_type_i)' failed at src/core/src/op/einsum.cpp:185:
While validating node 'opset7::Einsum Einsum_117987 (opset8::Gather __module.pos_embed/aten::select/Gather[0]:f32[?,?], opset1::Multiply __module.pos_embed/aten::mul/Multiply[0]:f64[8]) -&gt; (dynamic[...])' with friendly_name 'Einsum_117987':
Inputs to Einsum operation must have the same type.

-- No conversion rule found for operations: aten::einsum, prim::TupleConstruct, prim::TupleUnpack
-- Conversion is failed for: prim::Constant, prim::ListConstruct
</pre><p> </p><p> </p><p>Text encoders and VAE are converted successfully. I use torch torch-2.4.0+cpu and torch-2.3.0+cpu versions, nncf 2.12 , openvino 2024.4.0.dev20240812 . May it be because of version of smth ?</p><p> </p><p>It also did not work without pip install sentencepiece protobuf , so maybe it would be good to add it to prerequisites</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/2285/discussion/#comment-1715641083'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

eaidova: thank you for remainder, I made patch in diffusers during testing, but forgot to apply it in notebook code

PR #2310, opened 2024-08-21
add phi3 agent support

openvino-dev-samples:


eaidova:
@openvino-dev-samples  as phi-3 now used as default, please remove test replacement, phi-3 should be suitable for testing

eaidova (approved):

PR #2314, opened 2024-08-22
NPU hello world notebooks compatibility issue workaround

raymondlo84:
added a quick workaround to install 2024.1 for the NPU demo. We will relax this once 2024.4 addressed this issue.


eaidova:
@raymondlo84 thanks, but I prefer another solution that should be invariant to openvino version. As it is tutorial in hello-world series, we do not demonstrate model conversion in them, so it should be better to use preconverted model instead of pinning openvino https://github.com/openvinotoolkit/openvino_notebooks/pull/2316

raymondlo84:
Thanks!

PR #2326, opened 2024-08-27
Updating default webcam resolution 

raymondlo84:
Adding resolution config so we will use HD resolution images for webcam demo.


utils/notebook_utils.py
```diff
@@ -252,6 +252,10 @@ def __init__(self, source, size=None, flip=False, fps=None, skip_first_frames=0)
 
         self.cv2 = cv2  # This is done to access the package in class methods
         self.__cap = cv2.VideoCapture(source)
+        #try HD by default to get better video quality 
+        self.__cap.set(cv2.CAP_PROP_FRAME_WIDTH, 1280)
```
eaidova: should it be configurable with default values? What if user want different resolution or use different webcam?

raymondlo84: Good point. Updated.

raymondlo84 (commented):
Like this? :)

utils/notebook_utils.py
raymondlo84: https://github.com/openvinotoolkit/openvino_notebooks/pull/2326/commits/d1967a746ec76214b3f6194fd80d2f1219fd4243 =)

eaidova (approved):

PR #2300, opened 2024-08-16
Fix a problem with silero downloading

sbalandi:
[CVS-149572](https://jira.devtools.intel.com/browse/CVS-149572)

PR #2359, opened 2024-09-03
Move .cache/torch/hub to local for openvoice

sbalandi:
CVS-149572


sbalandi:
previous fix [PR#2300](https://github.com/openvinotoolkit/openvino_notebooks/pull/2300) based on issue https://github.com/myshell-ai/OpenVoice/issues/57 and [solution](https://github.com/myshell-ai/OpenVoice/issues/57#issuecomment-1913631053) / https://github.com/myshell-ai/OpenVoice/blob/main/docs/QA.md#issues-with-installation

here TORCH_HOME is moved to local path, it should be absolute because the path is used to creation simulink inside the se_extractor.get_se

eaidova:
@sbalandi  can you please also setting cache dir tgenerally in validation script too like we do for hf hub https://github.com/openvinotoolkit/openvino_notebooks/blob/latest/.ci/validate_notebooks.py#L203-L204

I think it can provide better cache controlling and be beneficial for other notebooks too (e.g. automatic cleaning disk space after notebook run)

eaidova (approved):

PR #2374, opened 2024-09-10
Fix typo of variable `embed_tokem` to `embed_token` in ov_phi3_vision.py at `line 381`

szeyu:
the original note write self.embed_token which is not existed. self.embed_tokem should be used instead


eaidova:
@szeyu  thank you, but looks like it is typo in variable name, should be  in opposite`embed_token`

szeyu:
> @szeyu thank you, but looks like it is typo in variable name, should be in opposite`embed_token`

Hi @eaidova, thanks for your clarification that it is a typo of variable name. Now I have submitted another commit where typos are fixed

eaidova (approved):

PR #2422, opened 2024-09-29
switch the order of llamaindex installation

openvino-dev-samples:
there are dependencies in llama-index-llms-openvino installation

[llama-index-llms-openvino](https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/llms/llama-index-llms-openvino/pyproject.toml#L37)


eaidova:
@openvino-dev-samples I made this intentionally to install llama-index openvino without dependencies because this notebook forces reinstalling openvino on 2024.3

openvino-dev-samples:
@eaidova updated

eaidova (approved):

PR #2429, opened 2024-10-01
fix win

sbalandi:
Task: [CVS-153935](https://jira.devtools.intel.com/browse/CVS-153935)


eaidova:
@sbalandi just as an idea, maybe we can move notebooks in different directories? notebooks/sam2-image-segmentation and notebooks/sam2-video-segmentation? I think it should simplify many things (it does not required these proposed workarounds) and also aligned with our desire to have one notebook per one directory

eaidova (approved):

PR #2440, opened 2024-10-10
Update gradio_helper.py to fix processor input issues for mllama-3.2

zhuo-yoyowz:
Run notebook and met the same issue as described in https://github.com/meta-llama/llama-recipes/issues/695.
Fixed the issue using fix here https://github.com/meta-llama/llama-recipes/pull/696/files.

### DIFF
--- a/notebooks/mllama-3.2/gradio_helper.py
+++ b/notebooks/mllama-3.2/gradio_helper.py
@@ -68,7 +68,7 @@ def make_demo(model, processor):
         prompt = processor.tokenizer.apply_chat_template(conversation, tokenize=False, add_generation_prompt=True)
         print(f"prompt is -\n{prompt}")
         image = Image.open(image)
-        inputs = processor(image, prompt, return_tensors="pt")
+        inputs = processor(images=image, text=prompt, return_tensors="pt")
 
         streamer = TextIteratorStreamer(
             processor,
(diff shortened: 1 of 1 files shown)


eaidova:
Thanks, @zhuo-yoyowz could you please provide info about transformers version which you use? I believe these changes may be unsafe and nonbackward compatible as on my end gradio worked well on transformers 4.45.0

notebooks/mllama-3.2/gradio_helper.py
```diff
@@ -68,7 +68,7 @@ def bot_streaming(message, history):
         prompt = processor.tokenizer.apply_chat_template(conversation, tokenize=False, add_generation_prompt=True)
         print(f"prompt is -\n{prompt}")
         image = Image.open(image)
-        inputs = processor(prompt, image, return_tensors="pt")
+        inputs = processor(image, prompt, return_tensors="pt")
```
eaidova: I suppose this fix will be more reliable:
```suggestion
        inputs = processor(images=image, text=prompt, return_tensors="pt")
```

zhuo-yoyowz:
> Thanks, @zhuo-yoyowz could you please provide info about transformers version which you use? I believe these changes may be unsafe and nonbackward compatible as on my end gradio worked well on transformers 4.45.0

I used transformers 4.45.1, but I also tried 4.45.0 but failed. Not sure where the issue comes from.

notebooks/mllama-3.2/gradio_helper.py
zhuo-yoyowz: > I suppose this fix will be more reliable:

Agree. This one looks good. Thank you!

eaidova (approved):

PR #2446, opened 2024-10-14
Update hello-world.ipynb 

raymondlo84:
removing the matplotlib <3.7 because it will conflict with numpy 2.x on Windows.

### DIFF
--- a/notebooks/hello-world/hello-world.ipynb
+++ b/notebooks/hello-world/hello-world.ipynb
@@ -43,7 +43,7 @@
     "import platform\n",
     "\n",
     "# Install openvino package\n",
-    "%pip install -q \"openvino>=2023.1.0\" opencv-python tqdm\n",
+    "%pip install -q \"openvino>=2023.1.0\" opencv-python tqdm \"matplotlib>=3.4\"\n",
     "\n"
    ]
   },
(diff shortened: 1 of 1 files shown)


eaidova:
@raymondlo84  thank you for you patch, but it looks not fully correct. notebook uses matplotlib for visualization, if you fully remove its installation from notebook, notebook will fail (and not only for windows and numpy 2.0), so please keep matplotlib installation in notebook, but remove upperbound for windows

notebooks/hello-world/hello-world.ipynb
```diff
@@ -44,11 +44,6 @@
     "\n",
     "# Install openvino package\n",
     "%pip install -q \"openvino>=2023.1.0\" opencv-python tqdm\n",
```
eaidova:
```suggestion
    "%pip install -q \"openvino>=2023.1.0\" opencv-python tqdm \"matplotlib>=3.4\"\n",
```

raymondlo84:
> @raymondlo84 thank you for you patch, but it looks not fully correct. notebook uses matplotlib for visualization, if you fully remove its installation from notebook, notebook will fail (and not only for windows and numpy 2.0), so please keep matplotlib installation in notebook, but remove upperbound for windows

Good catch. I thought it was in the requirements.txt before, and thanks!!

eaidova (approved):

PR #2453, opened 2024-10-16
Catvton

aleksandr-mokrov:
CVS-154070


notebooks/catvton/README.md
```diff
@@ -0,0 +1,29 @@
+# Virtual Try-On with CatVTON and OpenVINO
+
+### Abstract
```
eaidova, line 3: for me it is slightly strange to start notebook and readme with abstract subarticle, that is stylistically different from other notebooks. Could you please remove header in both readme and notebook?

notebooks/catvton/catvton.ipynb
```diff
@@ -0,0 +1,261 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "a1f0cac2-0661-4974-a5c6-34d0bed7c92c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%pip install -q \"openvino>=2024.4\"\n",
+    "%pip install -q \"torch>=2.1\" \"diffusers>=0.29.1\" torchvision opencv_python \"numpy<2\" --extra-index-url https://download.pytorch.org/whl/cpu\n",
```
eaidova: does numpy < 2.0 is strict requirement or it is os-specific?

notebooks/catvton/catvton.ipynb
```diff
@@ -0,0 +1,261 @@
...
+    "%pip install -q fvcore \"pillow\" \"tqdm\" \"gradio>=4.36\" \"omegaconf==2.4.0.dev3\" av pycocotools cloudpickle scipy accelerate \"transformers>=4.27.3\""
```
eaidova, line 65: you said that notebook demostrates how to optimize, but I do not see nncf in requirements (under optimization we usually means quantization). Is there is some plans for that or optimization should be removed from description until this is not added?

scipy package is heavily depends from numpy, could you please check that it will not breaks if you do  not specify lower bound for scipy and have it for numpy?

notebooks/catvton/gradio_helper.py
```diff
@@ -0,0 +1,199 @@
...
+                        )
+                        condition_person_exm = gr.Examples(
+                            examples=[os.path.join(root_path, "condition", "person", _) for _ in os.listdir(os.path.join(root_path, "condition", "person"))],
+                            examples_per_page=4,
+                            inputs=cloth_image,
+                            label="Condition Reference Person Examples",
+                        )
+                        gr.Markdown('<span style="color: #808080; font-size: small;">*Condition examples come from the Internet. </span>')
+
+            image_path.change(person_example_fn, inputs=image_path, outputs=person_image)
```
eaidova: why this is needed?

notebooks/catvton/ov_catvton_helper.py
```diff
@@ -0,0 +1,209 @@
+import os
+from collections import namedtuple
+from pathlib import Path
+import warnings
+
+from detectron2.export import TracingAdapter
```
eaidova: I do not see anything detectrron2 related in dependencies, from what these imports comes from?

notebooks/catvton/ov_catvton_helper.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def forward(self, sample=None, timestep=None, encoder_hidden_states=None, return_dict=None):
+        result = self.unet(sample=sample, timestep=timestep, encoder_hidden_states=encoder_hidden_states, return_dict=False)
+        return result
+
+
+def download_models():
+    resume_path = "zhengchong/CatVTON"
+    base_model_path = "booksforcharlie/stable-diffusion-inpainting"
+    repo_path = snapshot_download(repo_id=resume_path)
```
eaidova: does original pipeline does not support model ids?

Possibly it is better to specify local_dir argument to keep downloaded files in notebook directory

utils/cmd_helper.py
```diff
@@ -0,0 +1,18 @@
...
+    repo_path = Path(repo_url.split("/")[-1].split(".")[0])
+
+    if not repo_path.exists():
+        exit_code = os.system(f"git clone {repo_url}")
+        if exit_code != 0:
+            raise Exception("Failed to clone the repository!")
+        if revision:
+            subprocess.Popen(["git", "checkout", revision], cwd=str(repo_path))
+
+    sys.path.insert(0, str(repo_path))
```
eaidova: I think this should be optional or moved outside cloning function, not always we clone repo for adding that to path.
also it should be good if cloning function will returns path to cloned repo

utils/cmd_helper.py
```diff
@@ -0,0 +1,18 @@
+import os
+import sys
+from pathlib import Path
+
+
+def clone_repo(repo_url: str, revision: str = None):
+    import subprocess  # nosec - disable B404:import-subprocess check
```
eaidova: why subprocess imported inside function while other libs outside?

it is part of standard lib, we usually moving imports inside only for some optional dependencies to prevent requirements to install it if some different function used that does not require this lib

utils/cmd_helper.py
```diff
@@ -0,0 +1,18 @@
...
+
+    repo_path = Path(repo_url.split("/")[-1].split(".")[0])
```
eaidova: what if repo name contains .?
e.g. https://github.com/openvinotoolkit/openvino.genai

notebooks/catvton/ov_catvton_helper.py
```diff
@@ -0,0 +1,209 @@
...
+        torch.jit._state._clear_class_state()
+
+
+class VaeEncoder(torch.nn.Module):
+    def __init__(self, vae):
+        super().__init__()
+        self.vae = vae
+
+    def forward(self, x):
+        return self.vae.encode(x).latent_dist.sample()
```
eaidova: sample() includes random uniform that can make reproducing pytorch results numerical unstable. Possbly better to return distribution and implement sample in model inference wrapper.

Example how it can be achieved: 
conversion https://github.com/huggingface/optimum/blob/main/optimum/exporters/utils.py#L124
inference https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_diffusion.py#L938

DiagonalGaussianDistribution is original vae encoder output that has sample() function

notebooks/catvton/catvton.ipynb
```diff
@@ -0,0 +1,261 @@
+{
```
eaidova, line 1: <p>&gt;to save it as XML file.</p><p> </p><p> </p><p> actually, openvino IR is not just XML, XML contains only model graph and operations definitions while weights stored in bin file. Such comments may mislead newcomers (e.g. in my practice, I sow cases when people really thought that only XML is enough for inference and try to run model after remove bin file ), could you please rephrase avoiding mention xml</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/2453/discussion/#comment-1811894008'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

eaidova, line 1: <pre class="ql-syntax" spellcheck="false">Line #16.    convert_pipeline_models(pipeline, VAE_ENCODER_PATH, VAE_DECODER_PATH, UNET_PATH)
</pre><p>could you please provide opportunity to see conversion code? in description above you speak about convert_model, but there is no usage it in notebook. Please explain that for simplifying conversion process you use helper function that is based on convert_model</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/2453/discussion/#comment-1811894011'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='6'/>

utils/cmd_helper.py
```diff
@@ -1,18 +1,17 @@
...
 import sys
 from pathlib import Path
 
 
-def clone_repo(repo_url: str, revision: str = None):
-    import subprocess  # nosec - disable B404:import-subprocess check
-
-    repo_path = Path(repo_url.split("/")[-1].split(".")[0])
+def clone_repo(repo_url: str, revision: str = None, add_to_sys_path: bool = True):
+    repo_path = Path(repo_url.split("/")[-1].replace(".git", ""))
```
eaidova, line 7: still question what about return repo_path as output?

utils/cmd_helper.py
```diff
@@ -1,18 +1,17 @@
...
 
     if not repo_path.exists():
         exit_code = os.system(f"git clone {repo_url}")
```
eaidova: discussed several times on meetings, please do nit use os.system

eaidova (approved):

notebooks/catvton/catvton.ipynb
```diff
@@ -0,0 +1,271 @@
...
```
eaidova, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.        %pip install -q "numpy&lt;2.0.0"
</pre><p> this is not critical, but it is better to install numpy in the last order to guarantee that exact this version will be used in notebook (some of third-party may override that)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/2453/discussion/#comment-1815035105'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

PR #2456, opened 2024-10-17
Stable Diffusion V2 with torch FX

anzr299:
Created a new example notebook which Converts the unet model to an Fx graph and quantized it.

Reopens #2289


anzr299:
Requesting review from @MaximProshin @alexsu52 @daniil-lyakhov

PR #2494, opened 2024-10-30
reduces memory consumption for qwen2vl

eaidova:


openvino-dev-samples:
I got this error during model conversation with this patch:
image

eaidova:
@openvino-dev-samples this is not related to patch, this is related to changes in transformers 4.46.0. I have fix on ov side that should be available in the latest openvino nightly(https://github.com/openvinotoolkit/openvino/pull/27315) or alternatively, you can downgrade transformers to 4.45 or upgrade on 4.46.1

PR #2539, opened 2024-11-19
instant-id: added deleting of full precision pipeline before quantization

nikita-savelyevv:
**Changes**:
- In order to reduce peak memory footprint, full precision OV pipeline is now deleted right after calibration dataset is collected. This way it does not take up additional memory during quantization.
- Comparing inference speed of original and optimized pipelines is now optional (disabled by default) for the same reason.

After applying the changes and updating to openvino-nightly, the peak memory is observed to drop from 120 GB to 60 GB.

Related ticket: 146016

### DIFF
--- a/notebooks/instant-id/instant-id.ipynb
+++ b/notebooks/instant-id/instant-id.ipynb
@@ -1811,43 +1811,6 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": 22,
-   "id": "e111bc7e-fac9-4881-a3d2-f39057f369b6",
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "text_encoder = core.compile_model(ov_text_encoder_path, device.value)\n",
-    "text_encoder_2 = core.compile_model(ov_text_encoder_2_path, device.value)\n",
-    "vae_decoder = core.compile_model(ov_vae_decoder_path, device.value)\n",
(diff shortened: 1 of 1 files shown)


nikita-savelyevv:
@l-bat , it may be beneficial to consider logic like this when implementing quantization of large pipelines in the future

eaidova (approved):

eaidova:
@nikita-savelyevv please fix code formatting

notebooks/instant-id/instant-id.ipynb
```diff
@@ -1810,43 +1810,6 @@
     "device"
```
eaidova, line 1810: <pre class="ql-syntax" spellcheck="false">Line #4.    def create_ov_pipe():
</pre><p>maybe better to allow specify paths for reusing the same way of pipeline creation for int8 pipeline too?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/2539/discussion/#comment-1849850461'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='36'/>

nikita-savelyevv: Done

PR #2558, opened 2024-11-25
Cleanup models after conversion

aleksandr-mokrov:
CVS-157654

### DIFF
--- a/notebooks/catvton/catvton.ipynb
+++ b/notebooks/catvton/catvton.ipynb
@@ -240,13 +240,12 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 8,
+   "execution_count": null,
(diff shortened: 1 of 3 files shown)


eaidova:
thanks, but @aleksandr-mokrov that  is unlikely solve issue provided in the ticket.

OpenVINO uses mmap for reduce amount of required memory for model loading. This approach brings own limitations on windows - if file in use, OS blocks access to it until usage will be not finished (process finished itself or object explicitly removed by user).

Current problem that in this notebook we have instance of fp pipeline with compiled models and then try to read them again for quantization. So to solve problem, you need to remove ov_pipeline before start quantization.

eaidova (approved):

PR #2588, opened 2024-12-10
Fix load_image call

aleksandr-mokrov:
CVS-159041


sbalandi (approved):

sbalandi:
Did it work before ? If yes, shouldn't the lower version of diffusers be changed ?

aleksandr-mokrov:
> Did it work before ? If yes, shouldn't the lower version of diffusers be changed ?

It is not because of diffusers. Recently we changed our load_image and that changes affected this function call, but it is load_image from diffusers not our.

sbalandi:
> > Did it work before ? If yes, shouldn't the lower version of diffusers be changed ?
> 
> It is not because of diffusers. Recently we changed our load_image and that changes affected this function call, but it is load_image from diffusers not our.

if it is not our function, how could changes in ours affect it? it was used our before ?

aleksandr-mokrov:
> > > Did it work before ? If yes, shouldn't the lower version of diffusers be changed ?
> > 
> > 
> > It is not because of diffusers. Recently we changed our load_image and that changes affected this function call, but it is load_image from diffusers not our.
> 
> if it is not our function, how could changes in ours affect it? it was used our before ?

The call of this function was changed too by mistake

andrei-kochin (approved):

PR #2593, opened 2024-12-12
[ARM][CPU] Update Ultralytics version for openvino 2024.5

allnes:
Ultralytics in https://github.com/ultralytics/ultralytics/pull/16600 updated version of OpenVINO runtime
Latest version of openvino 2024.0 in Ultralytics, increasing version of Ultralytics helps to speed up inference on arm 

CVS-160346

### DIFF
--- a/notebooks/fast-segment-anything/fast-segment-anything.ipynb
+++ b/notebooks/fast-segment-anything/fast-segment-anything.ipynb
@@ -64,7 +64,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%pip install -q \"ultralytics==8.2.24\" \"matplotlib>=3.4\" \"onnx<1.16.2\" tqdm --extra-index-url https://download.pytorch.org/whl/cpu\n",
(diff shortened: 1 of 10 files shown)


eaidova:
@allnes thaqnks, but looks like it requires more changes as ultralytics API changed and incompatible with code used in notebooks. @aleksandr-mokrov could you please check?

allnes:
> @allnes thaqnks, but looks like it requires more changes as ultralytics API changed and incompatible with code used in notebooks. @aleksandr-mokrov could you please check?

I can only make changes where it didn't affect the API, but alas then the old version of Ultralitycs, will only load OpenVINO 2024.0, @eaidova please write, what I can change this, thank you

eaidova:
> > @allnes thaqnks, but looks like it requires more changes as ultralytics API changed and incompatible with code used in notebooks. @aleksandr-mokrov could you please check?
> 
> I can only make changes where it didn't affect the API, but alas then the old version of Ultralitycs, will only load OpenVINO 2024.0, @eaidova please write, what I can change this, thank you

@allnes do not worry, we will handle it by ourself (that is why I asked @aleksandr-mokrov  to take a look :))

PR #2603, opened 2024-12-17
Add macos arm64 compatibility

aleksandr-mokrov:
CVS-159110


notebooks/jax-to-openvino/jax-classification-to-openvino.ipynb
```diff
@@ -99,7 +99,10 @@
    "outputs": [],
    "source": [
     "%pip install -q \"openvino>=2024.5.0\"\n",
```
eaidova: probably need to update openvino on nightly to resolve unsupported op issue (or some specific jax version is required)

eaidova (approved):

PR #2714, opened 2025-02-04
Update to genai usage

aleksandr-mokrov:
CVS-161647


aleksandr-mokrov:
tiny-sd needs to be excluded for awhile

PR #2734, opened 2025-02-09
add batch speedup evaluation to fastdraft_deepseek notebook

shira-g:
Added a section to fastdraft_deepseek notebook to evaluate the speedup of deepseek with speculative decoding over multiple examples


shira-g:
@eaidova @ofirzaf

eaidova (approved):

PR #2732, opened 2025-02-07
deepseek-r1 Notebook: Add Option for DeepSeek-R1-Distill-Qwen-32B Model

ekurniaw:
- https://github.com/openvinotoolkit/openvino_notebooks/issues/2718
 - The 32B is the one comparable in capability as OpenAI-o1-mini 100B.
 - Tested with OpenVINO 2024.6 on ARL-S 285K CPU with 32 GB RAM and 500 GB drive (200 GB for vRAM swap file and 200 GB to store the model). Also, tested successfully on ARL-H 285H iGPU with 64 GB RAM.
 - Tested with OpenVINO 2025.0 on ARL-S 285K CPU as well.

PR #2737, opened 2025-02-10
deepseek-r1 Notebook: Add Option for DeepSeek-R1-Distill-Qwen-32B Model

ekurniaw:


ekurniaw:
Combined with https://github.com/openvinotoolkit/openvino_notebooks/pull/2732 instead.

PR #2749, opened 2025-02-13
use latest rolling gpu driver to support latest platforms

dtrawins:


ericwadams:
I tried this on a Dell Lunar Lake laptop and the GPU showed up using Podman on WSL when trying the gpu-devices.ipynb notebook.

`podman run -it --net=host --privileged --device /dev/dxg --volume /usr/lib/wsl:/usr/lib/wsl -p 8888:8888 openvino_notebooks_pull2749`

From within the notebook

```
import openvino as ov

core = ov.Core()
core.available_devices
['CPU', 'GPU']
```

eaidova (approved):

PR #2796, opened 2025-03-04
Incorrect chat template caused Unexpected outputs from r1 distill model

shrey2003:
Corrected Deepseek Distill chat templates


shrey2003:
@eaidova This is a small PR correcting the chat template of deepseek distill which I found while going through the code  please check

eaidova (approved):

PR #2798, opened 2025-03-04
Keras with OpenVINO backend 

aleksandr-mokrov:
CVS-159587


notebooks/keras-with-openvino-backend/keras-with-openvino-backend.ipynb
```diff
@@ -0,0 +1,339 @@
+{
```
eaidova, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    %pip install -q keras keras-hub
</pre><p>openvino supported only for keras&gt;=3.8, I think it is better to specify lower bound explicitly</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/2798/discussion/#comment-1987037964'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='2'/>

eaidova, line 1: <p>I think it is good to mention how these data can be obtained (using tokenizer?)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/2798/discussion/#comment-1987037967'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='10'/>

.ci/skipped_notebooks.yml
```diff
@@ -504,4 +504,10 @@
...
         - ubuntu-22.04
-        - windows-2019
\ No newline at end of file
+        - windows-2019
+- notebook: notebooks/keras-with-openvino-backend/keras-with-openvino-backend.ipynb
+  skips:
+    - os:
+        - windows-2019
+    - python:
+        - '3.12'
```
eaidova, line 513: is it related to your comment about tensorflow-text? I think it will be more informative if you add recommendation to downgrade python to 3.11 on windows if tokenizer does not work in notebook

notebooks/keras-with-openvino-backend/keras-with-openvino-backend.ipynb
eaidova, line 1: <p>I think a little bit more context required here:</p><p> </p><p>Keras provides list of pretrained for general purposes models that can be used for fine-tuning on specific task.</p><p>We will use bert model using <a href="https://keras.io/keras_hub/api/base_classes/text_classifier/#textclassifier-class" target="_blank">TextClassificator</a> class. OpenVINO API provides only inference capabilities, it means that before moving to openvino backend, you need to train model on own data used one of backends that supports training and next, when your model training process is finished, you can move to openvino for getting inference speed up. Here general steps that you need for that:</p><ol><li>specify backend using environment variable</li><li>create model instance</li><li>run model prediction</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/2798/discussion/#comment-1987052387'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='6'/>

.ci/skipped_notebooks.yml
aleksandr-mokrov: No. tensorflow-text doesn't provide package for macOS x86-64 since 2.16.1 version. And this version doesn't have package for python 3.12 and for macOS ARM.

eaidova (approved):

PR #2814, opened 2025-03-13
Fix: SAM2 Video Tracking 

RicoSuaveGuapo:
## Scope

Fix for tracking fail when max_frame_size is larger than around 60 frames. Example of original result

image

You can easily reproduce the above result by modifying 
```python
segmenter_video = SegmenterVideo(ov_predictor_video, max_frame_size=200)
```
in the original tutorial jupyter notebook 

## Implementation

The fix is simple. 
1. SAM2 only recognizes the point input with padding for the point prompt.
2. After the 0th frame (the conditional frame), there will be no point_input anymore, one has to add an empty point into.

## Result

Before

https://github.com/user-attachments/assets/4e2efb3b-3c04-4486-ab23-1333ba2d3ae8

After

https://github.com/user-attachments/assets/92ac588b-3e61-41b0-bec7-276a6271c67c

### DIFF
--- a/notebooks/sam2-video-segmentation/ov_sam2_helper.py
+++ b/notebooks/sam2-video-segmentation/ov_sam2_helper.py
@@ -70,6 +70,16 @@ class SamVideoMaskPredictorModel(torch.nn.Module):
 
     def _embed_points(self, point_coords: torch.Tensor, point_labels: torch.Tensor, pad: bool = True) -> torch.Tensor:
         point_coords = point_coords + 0.5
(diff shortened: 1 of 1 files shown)


RicoSuaveGuapo:
Thank you all for porting the model—I understand this was a huge undertaking.
I myself also spent significant time understanding the differences between the PyTorch implementation and your method, even though the fix only take few lines.

I greatly appreciate your taking the time to review this. Thank you in advance!

eaidova (approved):

PR #2832, opened 2025-03-19
Add prompt-lookup-decoding.ipynb notebook

sbalandi:
[CVS-164270](https://jira.devtools.intel.com/browse/CVS-164270)


eaidova:
@sbalandi could you please provide output logs with speedup comparision in notebook (now only speculative decoding part has output, other sections are not.

Another question, what is the reason to skip model on ci, does it too large to be executed in github actions? If yes, it is possible to use something smaller for testing only (e.g. tiny-llama)

eaidova (approved):

PR #2977, opened 2025-06-06
add minicpm4 to llm chatbot

openvino-dev-samples:
Please note this model is very sensitive to the sampling parameters, so i suggest to use greedy search or a small top_k value for sampling. Even the official example has same issue:https://huggingface.co/openbmb/MiniCPM4-8B#inference-with-transformers


AnishaUdayakumar (approved):
Tested the changes for MiniCPM4 and Tinyllama and the chatbot inference works well.

notebooks/llm-chatbot/gradio_helper.py
```diff
@@ -110,7 +110,7 @@ def make_demo(run_fn: Callable, stop_fn: Callable, title: str = "OpenVINO Chatbo
                         with gr.Row():
                             top_k = gr.Slider(
                                 label="Top-k",
-                                value=50,
+                                value=1,
```
aleksandr-mokrov, line 113: This change affects all models in this notebook, most of them have default value 50. Adding a new model shouldn't affect other models, or we should check this change for all models.

openvino-dev-samples: greedy search should work for all of models

sbalandi: Did you check it ? Both model work notrmal with tok-k 50 for me (more precisely, the same as with 1).
But I think it's ok to set greedy(), but for that, please, set also temperature = 0 and add to description of this parameter, that if it's 0, then generation will be run in greedy mode/woth do_sample=False. 
Also, please, make `minimum` for Top-p 1.

openvino-dev-samples: Yes, both @AnishaUdayakumar and I verified it. Setting the default sampling method as greedy search is the most safe solution for all of LLMs.  I think Topk=1 will trigger greedy mode as well ?

sbalandi: if temperature > 0 , do_sample will set to True , it is multinomial semantically 

What did you get ? Did you get that the output contains garbage with Top k 50 and don't have it with 1 ?

openvino-dev-samples: “output contains garbage with Top k 50 and don't have it with 1”， yes, exactly.

openvino-dev-samples: > if temperature > 0 , do_sample will set to True , it is multinomial semantically
> 
> What did you get ? Did you get that the output contains garbage with Top k 50 and don't have it with 1 ?

So i think topk=1 is a safe number, and still working for `do_sampling=True`.

sbalandi: why we need do_sampling in that case ?

sbalandi (changes requested):
please, make `minimum` for Top-p  1
please, make temperature 0

sbalandi:
> please, make `minimum` for Top-p 1 please, make temperature 0

Top p to 0.1 or 0.01, not 1, multinomial sampling will fail with top-p 0 in genAI https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/generation_config.cpp#L225

openvino-dev-samples:
> please, make `minimum` for Top-p 1 please, make temperature 0

done

notebooks/llm-chatbot/gradio_helper.py
```diff
@@ -96,7 +96,7 @@ def make_demo(run_fn: Callable, stop_fn: Callable, title: str = "OpenVINO Chatbo
                         with gr.Row():
                             top_p = gr.Slider(
                                 label="Top-p (nucleus sampling)",
-                                value=1.0,
+                                value=0.1,
                                 minimum=0.0,
```
sbalandi, line 100: value=1.0,
minimum=0.01,

notebooks/llm-chatbot/gradio_helper_genai.py
```diff
@@ -207,7 +207,7 @@ def stop_chat_and_clear_history(streamer):
                         with gr.Row():
                             top_p = gr.Slider(
                                 label="Top-p (nucleus sampling)",
-                                value=1.0,
+                                value=0.1,
                                 minimum=0.0,
```
sbalandi, line 211: value=1.0,
minimum=0.01,

notebooks/llm-chatbot/gradio_helper.py
sbalandi:
```suggestion
                                value=1.0,
                                minimum=0.01,
```

notebooks/llm-chatbot/gradio_helper_genai.py
sbalandi:
```suggestion
                                value=1.0,
                                minimum=0.01,
```

notebooks/llm-chatbot/gradio_helper.py
openvino-dev-samples: Sorry, its my mistake. I was sleepy when making these changes yesterday : ).

notebooks/llm-chatbot/gradio_helper_genai.py
```diff
@@ -91,8 +91,8 @@ def default_partial_text_processor(partial_text: str, new_text: str):
         updated text string
 
         """
-        new_text = re.sub(r"^<think>", "<em><small>I am thinking...", new_text)
-        new_text = re.sub("</think>", "I think I know the answer</small></em>", new_text)
+        # new_text = re.sub(r"^<think>", "<em><small>I am thinking...", new_text)
+        # new_text = re.sub("</think>", "I think I know the answer</small></em>", new_text)
```
sbalandi: Why is this commented out?

openvino-dev-samples: Another mistake

sbalandi (approved):

PR #3000, opened 2025-06-25
Add ACTIVATIONS_SCALE_FACTOR in compile config if GPU

aleksandr-mokrov:
CVS-168629


notebooks/stable-cascade-image-generation/stable-cascade-image-generation.ipynb
```diff
@@ -556,14 +556,15 @@
...
-   "execution_count": 13,
+   "execution_count": null,
    "id": "271b7d3a-6517-48d5-b04f-9104547cf9b8",
    "metadata": {},
    "outputs": [],
    "source": [
     "class PriorPriorWrapper:\n",
     "    def __init__(self, prior_path, device):\n",
-    "        self.prior = core.compile_model(prior_path, device.value)\n",
+    "        config = {\"ACTIVATIONS_SCALE_FACTOR\": \"128\"} if device.value == \"GPU\" else {}\n",
```
davidsnam-intel, line 566: Depending on the user's environment, there can be GPU more than one. In that case, `device.value` can be _GPU.0_, _GPU.1_. That's why I suggested `config = {"ACTIVATIONS_SCALE_FACTOR": "128"} if 'GPU' in device.value else {}`. However you can decide based on the policy of openvino_notebooks.

davidsnam-intel (approved):

PR #3012, opened 2025-06-30
Fix model download issue: Download both .xml and .bin files

ezelanza:
### What changed

This PR addresses an issue where the model downloader was only fetching the `.xml` (model architecture) file but not the corresponding `.bin` (model weights) file for OpenVINO models. This mismatch caused runtime errors like:

### Why this was needed

Both the `.xml` and `.bin` files are required for OpenVINO IR (Intermediate Representation) models to run inference. Without the weights file, the model can't load or perform any predictions. Users trying to run the **person detection** (`person-detection-0202`) or **person re-identification** (`person-reidentification-retail-0287`) pipelines were hitting this error.

### What models are impacted

- **person-detection-0202**
- **person-reidentification-retail-0287**

These models are often used together for multi-step pipelines (e.g., detect a person, then re-identify them across frames or cameras).

### What I did

- Updated the download logic to ensure **both** the `.xml` and `.bin` files for each model are downloaded and available at runtime.
- Ran end-to-end tests on both models to confirm they load successfully and perform inference without errors.
- Cleaned up leftover cell outputs in the notebook for readability and to prevent confusion when users open the notebook.

### How to test

1. Run the notebook end-to-end.
2. Confirm that both models load without errors.
3. Check that both `.xml` and `.bin` files exist in the expected download directory.

### Before this fix:

Running inference would fail with missing weights error.

### After this fix:

Both models load successfully and inference runs as expected.

---

### DIFF
--- a/notebooks/person-tracking-webcam/person-tracking.ipynb
+++ b/notebooks/person-tracking-webcam/person-tracking.ipynb
@@ -172,23 +172,34 @@
     "# A directory where the model will be downloaded.\n",
     "base_model_dir = \"model\"\n",
     "precision = \"FP16\"\n",
+    "\n",
     "# The name of the model from Open Model Zoo\n",
     "detection_model_name = \"person-detection-0202\"\n",
-    "download_det_model_url = (\n",
-    "    f\"https://storage.openvinotoolkit.org/repositories/open_model_zoo/2023.0/models_bin/1/{detection_model_name}/{precision}/{detection_model_name}.xml\"\n",
-    ")\n",
     "detection_model_path = Path(base_model_dir) / detection_model_name / precision / f\"{detection_model_name}.xml\"\n",
+    "detection_model_bin_path = Path(base_model_dir) / detection_model_name / precision / f\"{detection_model_name}.bin\"\n",
     "\n",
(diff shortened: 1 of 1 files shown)


notebooks/person-tracking-webcam/person-tracking.ipynb
```diff
@@ -172,23 +172,34 @@
     "# A directory where the model will be downloaded.\n",
```
aleksandr-mokrov, line 172: <pre class="ql-syntax" spellcheck="false">Line #15.        download_ir_model(download_det_model_xml_url, Path(base_model_dir) / detection_model_name / precision)
</pre><p>download_ir_model accepts xml url, and then download bin and xml files. It checks separately if each of them already exists and downloads if not. So, it is enough to remove<code>if not detection_model_path.exists():</code> and  <code>if not reidentification_model_path.exists():</code> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openvinotoolkit/openvino_notebooks/pull/3012/discussion/#comment-2180036507'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

PR #3271, opened 2026-01-28
Add Qwen3-30B-A3B to llm chatbot notebooks

aleksandr-mokrov:
CVS-174432


openvino-dev-samples:
Shall we add an option to download its IR model from HF ? As it will need huge memory to convert it locally.

aleksandr-mokrov:
> Shall we add an option to download its IR model from HF ? As it will need huge memory to convert it locally.

We have this option, but the IR model is not downloaded on HF yet.

PR #3352, opened 2026-03-17
Sqip quantizatio for GPU device

aleksandr-mokrov:
CVS-175369


brmarkus:
@aleksandr-mokrov can you add a note about _why_ quantization got skipped for the GPU device, please? Breaks functionality, introducing noise, reduce accuracy?

PR #3369, opened 2026-03-24
fix: replace deprecated openvino.runtime imports with openvino namespace in helper scripts

abhayjeet5465:
## Summary
Fixes deprecated `openvino.runtime` sub-module imports across 9 helper scripts.

## Problem
The `openvino.runtime` nested namespace was deprecated and is removed in 
OpenVINO 2026.0. Any script using `from openvino.runtime import ...` will 
throw an ImportError on the current release.

Related: https://github.com/openvinotoolkit/openvino/blob/master/docs/release_notes/index.md

## Changes
Updated the following 9 files:
- `notebooks/glm4-v/ov_glm4v_helper.py`
- `notebooks/janus-multimodal-generation/ov_janus_helper.py`
- `notebooks/minicpm-o-omnimodal-chatbot/minicpm_o_helper.py`
- `notebooks/minicpm-o-4.5/minicpm_o_4_5_helper.py`
- `notebooks/qwen2-audio/ov_qwen2_audio_helper.py`
- `notebooks/qwen2.5-omni-chatbot/qwen2_5_omni_helper.py`
- `notebooks/qwen3-asr/qwen_3_asr_helper.py`
- `notebooks/qwen3-tts/qwen_3_tts_helper.py`
- `notebooks/ct-segmentation-quantize/custom_segmentation.py`

**Fix applied in each file:**
```python
# Before (deprecated)
from openvino.runtime import opset13
from openvino.runtime import PartialShape

# After (correct)
from openvino import opset13
from openvino import PartialShape
```

## Testing
Verified zero remaining deprecated imports using:
```
Get-ChildItem -Path notebooks/ -Recurse -Include "*.py" | 
Select-String -Pattern "from openvino\.runtime import|import openvino\.runtime as"
```
Output: empty (all fixed)


aleksandr-mokrov:
Thank you for the contribution, but this PR doesn't actually replace the deprecated imports — it makes the except ImportError fallback identical to the try block two lines above, effectively turning the try/except into dead code.

The current pattern is intentional:

```
try:
    from openvino import opset13        # modern API
except ImportError:
    from openvino.runtime import opset13  # fallback for older versions
```

The openvino.runtime import in the except branch is a backward-compatibility fallback, not the primary import path. It only runs when the modern from openvino import ... is unavailable (i.e., on older OpenVINO versions).

After this PR, both branches would execute the same from openvino import opset13, so if it fails in try, it will also fail in except — removing the fallback entirely while keeping a meaningless try/except wrapper.

Closing as the change would reduce compatibility without any functional benefit.

PR #3383, opened 2026-03-27
fix: replace tqdm.notebook import with tqdm in helper scripts

abhayjeet5465:
## Summary

Replaces `from tqdm.notebook import tqdm` with `from tqdm import tqdm` 
in 4 helper scripts.

## Problem

`tqdm.notebook` is designed exclusively for Jupyter notebook cells — it 
renders an IPython widget. When the same import is used inside a plain 
`.py` helper script (which gets imported by notebooks, not run as a cell), 
it either falls back silently or raises a rendering warning depending on 
the environment. The correct import for helper scripts is the standard 
`from tqdm import tqdm`, which works correctly in both notebook and 
non-notebook contexts.

## Files Changed

- `notebooks/catvton/catvton_quantization_helper.py`
- `notebooks/ltx-video/ltx_video_quantization_helper.py`  
- `notebooks/sam2-image-segmentation/automatic_mask_generation_helper.py`
- `notebooks/stable-diffusion-v3-torch-fx/sd3_quantization_helper.py`

## Verification
```powershell
Get-ChildItem -Path notebooks/ -Recurse -Include "*.py" |
  Select-String -Pattern "from tqdm\.notebook import tqdm"
```
Output: empty (all fixed)

### DIFF
--- a/notebooks/catvton/catvton_quantization_helper.py
+++ b/notebooks/catvton/catvton_quantization_helper.py
@@ -1,7 +1,7 @@
 from typing import Any
 from pathlib import Path
 
-from tqdm.notebook import tqdm
+from tqdm.auto import tqdm
 from transformers import set_seed
 import numpy as np
 import openvino as ov
--- a/notebooks/ltx-video/ltx_video_quantization_helper.py
+++ b/notebooks/ltx-video/ltx_video_quantization_helper.py
@@ -4,7 +4,7 @@ import numpy as np
 from pathlib import Path
 import openvino as ov
 import torch
-from tqdm.notebook import tqdm
+from tqdm.auto import tqdm
 
 negative_prompts = [
     "blurry unreal occluded",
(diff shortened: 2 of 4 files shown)


aleksandr-mokrov:
@abhayjeet5465 Thank you! The observation is fair — tqdm.notebook in .py helper files isn't ideal since they could be imported outside a notebook context.

However, replacing it with plain from tqdm import tqdm would lose the nice widget-style progress bars when these helpers are called from notebooks (which is their primary use case). The better replacement here is from tqdm.auto import tqdm — it auto-detects the environment and is already the dominant pattern in this repo (used in 7+ helper files like gradio_helper.py, ov_janus_helper.py, model_upcast_utils.py, etc.).

Could you update the PR to use tqdm.auto instead?

abhayjeet5465:
Thank you for the feedback @aleksandr-mokrov! I'll update the PR to use from tqdm.auto import tqdm instead. Making the change now.

abhayjeet5465:
Updated! All 4 files now use from tqdm.auto import tqdm instead of from tqdm.notebook import tqdm — it auto-detects the environment and matches the existing pattern across the repo. Thanks for the suggestion @aleksandr-mokrov!

aleksandr-mokrov:
@copilot resolve the merge conflicts in this pull request

PR #3404, opened 2026-04-07
[MiniCPMO-4.5]optimize the audio quality in streaming mode

openvino-dev-samples:
Add cache to the inputs of hift and estimator model for better audio quality in streaming mode.


aleksandr-mokrov:
The Token2wav group now contains 4 models (Flow Embeddings, Flow Encoder Chunk, Flow Estimator Chunk, HiFT), but total_models was not updated. And also description in readme and notebooks describes 15 models

aleksandr-mokrov (approved):

Issue #3412, opened 2026-04-12
[Bug]: ModuleNotFoundError: No module named 'blip_model' in BLIP Notebook


Shekar-77:
**Describe the bug**
The notebook fails with a ModuleNotFoundError: No module named 'blip_model' when attempting to import from a local file. This occurs because the utility script blip_model.py is missing from the local working directory.
Notebook ; [blip-visual-language-processing.ipynb](https://github.com/openvinotoolkit/openvino_notebooks/blob/latest/notebooks/blip-visual-language-processing/blip-visual-language-processing.ipynb)

**Steps to reproduce:**
Open the BLIP Visual Language Processing notebook.
Run the cell containing the import: from blip_model import OVBlipModel.
The cell fails with a ModuleNotFoundError

 **Expected behavior**
The notebook should either include blip_model.py in the repository structure or contain a cell that automatically downloads the necessary script if it is missing.
**Screenshots**

<img width="1071" height="442" alt="Image" src="https://github.com/user-attachments/assets/c1bf648c-c6ab-4833-bb40-68b18718d3ec" />

 **Installation instructions (Please mark the checkbox)**
[ . ] I followed the installation guide at https://github.com/openvinotoolkit/openvino_notebooks#-installation-guide to install the notebooks.

YuChern-Intel:
Closing issue as the pull request has been stalled, feel free to re-open or start a new issue if additional assistance is needed.

PR #3413, opened 2026-04-12
Fix: Add download check for blip_model.py utility script (Notebook: blip-visual-language-processing.ipynb)

Shekar-77:
Description:
This PR addresses the ModuleNotFoundError: No module named 'blip_model' error that occurs when the blip_model.py utility script is missing from the local working directory.
Notebook:https://github.com/openvinotoolkit/openvino_notebooks/blob/latest/notebooks/blip-visual-language-processing/blip-visual-language-processing.ipynb

Changes:
Just pushed 
```
if not Path("./blip_model.py").exists():
    download_file(url="https://raw.githubusercontent.com/openvinotoolkit/openvino_notebooks/latest/notebooks/blip-visual-language-processing/blip_model.py")
```
above the cell that requires blip_model.py file.

Resolves issue: #3412


aleksandr-mokrov:
Thank you for the contribution.
The issue is valid for the scenario where the notebook is downloaded as a standalone file without the rest of the repository — in that case `blip_model.py` would indeed be missing. However, when the repo is cloned following the [installation guide](https://github.com/openvinotoolkit/openvino_notebooks#-installation-guide), the file is always present.
That said, the PR doesn't fix the actual problem: **Cell 20** (`from blip_model import text_decoder_forward`) still runs **before** the new download-check cell, so the `ModuleNotFoundError` would still occur there. The fix also duplicates Cell 20's content in a later cell.
The correct fix would be to move the download check to **before** Cell 20 — just a cell reorder, no duplication needed.

The PR also includes unrelated upstream commits — please use a feature branch instead of `latest` for contributions.

Shekar-77:
Hey @aleksandr-mokrov I had created a feature branch that contains only the fix and not any other commits:
image
Branch link : https://github.com/Shekar-77/openvino_notebooks/tree/patch-1
But somehow made the pull request from the latest branch, however I have updated the cell placement on the latest branch. Do you want me to close this pr and make a pull request from the featured branch or will this pr work?

aleksandr-mokrov:
Thanks for the update. Yes, please close this PR and open a new one from your feature branch.

One important issue with the current state: commit 3d27315 removed the cell that patches text_decoder.forward, but that patch must run before generate_answer() — otherwise the model uses PyTorch inference instead of OpenVINO.

The correct fix is simply:

1. Add a new cell with the download check before the existing from blip_model import text_decoder_forward cell
2. Remove the duplicate download check from the OVBlipModel cell
3. Don't reorder anything else

(from pathlib import Path is already imported earlier in the notebook, so no need to add it again.)

PR #3414, opened 2026-04-14
[Qwen3-TTS/ASR]update quantization comments of model conversion

openvino-dev-samples:


notebooks/qwen3-tts/qwen_3_tts_helper.py
```diff
@@ -866,10 +866,10 @@ def forward_wrap_code_predictor(
         patch_stateful(ov_model, 2)
         print("✅ Talker Code Predictor model successfully converted")
 
-        if quantization_config is not None:
-            print(f"⌛ Weights compression with {quantization_config['mode']} mode started")
-            ov_model = nncf.compress_weights(ov_model, **quantization_config)
-            print("✅ Weights compression finished")
+        # if quantization_config is not None:
```
aleksandr-mokrov, line 869: Why was quantization disabled?

openvino-dev-samples: It is suggested by model vendor, as this part is sensitive to the accuracy.

aleksandr-mokrov: Then it should be explicitly. Now we have a comment in the notebook" # Set to {\"mode\": CompressWeightsMode.INT8_SYM} for INT8 quantization\n"", but should be something like # Note: weight compression is not supported for this model (Code Predictor is accuracy-sensitive). And here it is better to print that it is not supported and will be ignored.

notebooks/qwen3-asr/qwen3-asr.ipynb
```diff
@@ -260,7 +262,7 @@
     "convert_qwen3_asr_model(\n",
     "    model_id=model_id,\n",
     "    output_dir=ov_model_dir,\n",
-    "    quantization_config=None,  # Set to {\"mode\": \"INT8_SYM\"} for INT8 quantization\n",
+    "    quantization_config=None,  # Set to {\"mode\": CompressWeightsMode.INT8_SYM} for INT8 quantizationn\n",
```
aleksandr-mokrov: quantizationn - misspelling

aleksandr-mokrov (changes requested):

notebooks/qwen3-tts/qwen_3_tts_helper.py
openvino-dev-samples: Hi Only Code Predictor is not recommended to be quantized, but talker LLM can support quantization.

aleksandr-mokrov (approved):

PR #3442, opened 2026-05-06
Move VLM models to VLM unified notebook

aleksandr-mokrov:
Ticket: 184734


utils/llm_config.py
```diff
@@ -86,11 +86,41 @@ def lfm2_completion_to_prompt(completion):
             "supports_video": True,
             "exclude_on_devices": ["NPU"],
         },
+        "Qwen3-VL-2B-Instruct": {
```
openvino-dev-samples, line 84: I would suggest leaving Qwen3-VL-2B-Instruct as default model rather than Llava-Next-Video-7B, as its lighter and hotter in HF.

openvino-dev-samples (approved):

PR #3472, opened 2026-06-04
[ENH] YOLO2026 Advanced NNCF Quantization Techniques

tejaswinijayashanker943:
### Details:

This PR is to ensure that if basic nncf.quantize() gives an int8 model with accuracy drop> 1%, then use nncf with accuracy control.

### Tickets:
     N/A

### Expected results for accuracy-control models

The exact values depend on the validation subset and package versions, but the accuracy-control runs should keep the relative drop within 1 percent.

| Model | OpenVINO baseline mAP@50-95 | OpenVINO INT8 mAP@50-95 | Relative drop |
|---|---:|---:|---:|
| `yolo26s` | 0.4021 | about 0.3984 or higher | <= 1% |
| `yolo26m` | 0.4624 | about 0.4582 or higher | <= 1% |

### AI Assistance:

AI assistance used: No


aleksandr-mokrov:
Hi @tejaswinijayashanker943, thanks for the PR!

Could you please rebase your branch onto the latest upstream/latest? The notebooks/yolov26-optimization/ directory was added about 3 months ago and already contains several notebooks, including the yolov26-object-detection.ipynb you are adding. Your branch is based on a much older commit, so GitHub shows the file as a brand-new one (+1210/−0), while the real change against latest is only ~150 lines — and it will conflict with the existing file on merge.

Also, a question on the value: with the default quantization the notebook already keeps the accuracy drop within ~1% (as your own table shows). Since quantize_with_accuracy_control adds a fair amount of complexity (a labeled validation set + repeated validation), could you share the mAP comparison between plain default quantization and the accuracy-control path? If the difference is negligible, it may not be worth adding.

tejaswinijayashanker943:
> Hi @tejaswinijayashanker943, thanks for the PR!
> 
> Could you please rebase your branch onto the latest upstream/latest? The notebooks/yolov26-optimization/ directory was added about 3 months ago and already contains several notebooks, including the yolov26-object-detection.ipynb you are adding. Your branch is based on a much older commit, so GitHub shows the file as a brand-new one (+1210/−0), while the real change against latest is only ~150 lines — and it will conflict with the existing file on merge.
> 
> Also, a question on the value: with the default quantization the notebook already keeps the accuracy drop within ~1% (as your own table shows). Since quantize_with_accuracy_control adds a fair amount of complexity (a labeled validation set + repeated validation), could you share the mAP comparison between plain default quantization and the accuracy-control path? If the difference is negligible, it may not be worth adding.

Thank you @aleksandr-mokrov 
I will rebase with the latest master.

The original Notebooks work well with NANO variants. Still, with the Small, Medium, and Large variants, only _nncf.quantize()_ doesn't yield an OpenVINO model with a 1% drop; hence, I used various NNCF advanced quantization parameters.

Before NNCF with ACCURACY Control, only nncf.quantize gives us:


image

aleksandr-mokrov (approved):

PR #3542, opened 2026-07-08
add melo-tts notebook support

jiafeng5513:


aleksandr-mokrov:
Thanks @jiafeng5513! I don't think we need a separate melo-tts notebook — MeloTTS is already covered in notebooks/openvoice2-and-melotts/. This PR also vendors the entire external MeloTTS repo (157 files / ~284k lines, including duplicated code and large binary blobs like cmudict.pickle), which we can't take into the repository. Could you clarify what this adds over the existing notebook? Otherwise I'd suggest closing it.

PR #3548, opened 2026-07-10
Add MCNN model support on Intel HW

dhandhalyabhavik:
### What this PR do

This PR adds MCNN model support on Intel HW. MCNN is very old model (~7 years), it was written in tensorflow-1 so I had to first convert the model to pytorch and then OV format. So updated code for MCNN with model is available at https://github.com/dhandhalyabhavik/MCNN-ov

**Note:** The original link to model weights is not working, author provided link to a cloud system from which I had to manually download the model. The notebook uses model from the above forked repo. This was the only work around solution I could think of. It has all 3 weights original, ov-fp16 & ov-fp8.

CVS-190032

AI Disclaimer: 
This notebook is written by AI from above forked repo. All notebook cells are manually verified.


dhandhalyabhavik:
@aleksandr-mokrov please review

PR #3583, opened 2026-07-28
Totoka net error retest

totoka-intel:


.ci/validate_notebooks.py
```diff
@@ -28,6 +28,38 @@
 
 SEPARATED_VENV_NAME = Path("openvino_venv")
 
+NETWORK_ERROR_PATTERNS = [
```
akashchi, line 31: We have a [`workflow_rerunner`](https://github.com/openvinotoolkit/openvino/tree/master/.github/scripts/workflow_rerun) in the `openvino` repository. It acts similarly to this proposition, but works at the pipeline job level. I suggest reusing/adapting the [`log_analyzer.py`](https://github.com/openvinotoolkit/openvino/blob/master/.github/scripts/workflow_rerun/log_analyzer.py) part of the `workflow_rerunner`.

I think that it should improve error detection and make the code more maintainable.

PR #3629, opened 2026-08-27
Add pyannote-audio speaker diarization OpenVINO notebook

pravin25:
Adds a notebook enabling the pyannote speaker-diarization pipeline
(`pyannote/speaker-diarization-community-1` + `pyannote/segmentation-3.0`)
across CPU, Intel XPU, and OpenVINO (CPU/GPU) on the VoxConverse dataset.

Covers: OpenVINO IR export (FP16), single-file diarization per backend, latency
benchmarks, and VoxConverse DER scoring via an end-to-end `inference.ipynb`.


aleksandr-mokrov:
This can be a useful modern replacement for the speaker diarization notebook deprecated in 2024, using the newer `speaker-diarization-community-1` pipeline.
Please rework it into a shorter, cross-platform notebook following the repository structure and CI requirements. Full VoxConverse and XPU benchmarks should be optional.
Could you also check whether `annotation.get_timeline().support()` is the correct UEM? It may exclude false alarms during silence and make the reported 8.3% DER lower than expected.

aleksandr-mokrov (changes requested):
See my comment

pravin25:
> diarization notebook deprecated in 2024

ok will do

PR #3630, opened 2026-08-27
Add PyAnnote embedding OpenVINO implementation

pravin25:
Adds a notebook demonstrating OpenVINO enablement for the Hugging Face
[`pyannote/embedding`](https://huggingface.co/pyannote/embedding) speaker embedding model.

Covers: HF authentication, PyTorch → OpenVINO IR conversion, sample inference on
CPU/GPU, and the VoxCeleb1 EER benchmark.

Note: `pyannote/embedding` is a gated model — accept its terms and log in with a
Hugging Face token before running.


aleksandr-mokrov:
Thank you for the contribution. I do not recommend adding this as a standalone notebook.

The collection already had a [pyannote speaker diarization notebook](https://openvinotoolkit.github.io/openvino_notebooks/?search=pyannote ), which was deprecated in 2024. Rather than introducing another pyannote notebook based on the older `pyannote/embedding` model, the useful speaker embedding and verification scenario should be incorporated into PR #3629 as part of a modern replacement for the archived notebook.

pravin25:
> Thank you for the contribution. I do not recommend adding this as a standalone notebook.
> 
> The collection already had a [pyannote speaker diarization notebook](https://openvinotoolkit.github.io/openvino_notebooks/?search=pyannote), which was deprecated in 2024. Rather than introducing another pyannote notebook based on the older `pyannote/embedding` model, the useful speaker embedding and verification scenario should be incorporated into PR #3629 as part of a modern replacement for the archived notebook.

but this embedding model is different than PR#3629 which model using for embedding

PR #3646, opened 2026-09-10
Download assets from HF instead of Google Disc; allow `nltk` data download behind proxy in CI

akashchi:
Tickets: 158205, 192885, 192886.

### DIFF
--- a/.ci/skipped_notebooks.yml
+++ b/.ci/skipped_notebooks.yml
@@ -51,12 +51,6 @@
         - windows-2022
     - python:
         - '3.13'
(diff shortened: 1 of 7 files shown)


.ci/skipped_notebooks.yml
```diff
@@ -51,12 +51,6 @@
         - windows-2022
     - python:
         - '3.13'
-- notebook: notebooks/freevc-voice-conversion/freevc-voice-conversion.ipynb
```
akashchi, line 54: These two notebooks are passing; checked in one of the previous pipelines.

PR #3648, opened 2026-09-11
[notebooks] Fix OmniVoice audio encoder conversion broken by transformers 5.17

akashchi:
## Description

Fixes the `RuntimeError: Couldn't get TorchScript module by tracing` (caused by
`IndexError: Dimension out of range`) when converting the OmniVoice HiggsAudio
encoder to OpenVINO IR. The regression surfaced with the release of
`transformers==5.17`.

The patched HuBERT encoder forward in `omnivoice_helper.py` indexed the encoder
layer output with `[0]`, assuming a tuple return. As of `transformers` 5.17,
`HubertEncoderLayer.forward` returns a bare tensor, so `[0]` stripped the batch
dimension (`[1, seq, hidden]` → `[seq, hidden]`). The next layer's attention
then built a 3D tensor and failed at `key.transpose(2, 3)` during tracing.

## Fix

Removed the `[0]` so `hidden_states` keeps its full `[batch, seq, hidden]` shape.

## Ticket

194448

### DIFF
--- a/notebooks/omnivoice/omnivoice_helper.py
+++ b/notebooks/omnivoice/omnivoice_helper.py
@@ -284,7 +284,7 @@ def _convert_audio_tokenizer(
             for layer in self_.layers:
                 if output_hidden_states:
                     all_hidden = all_hidden + (hidden_states,)
-                # transformers>=5 HubertEncoderLayer returns a bare tensor, not a tuple
+                # transformers>=5.17 HubertEncoderLayer returns a bare tensor, not a tuple
                 hidden_states = layer(hidden_states, attention_mask=None)
             if output_hidden_states:
                 all_hidden = all_hidden + (hidden_states,)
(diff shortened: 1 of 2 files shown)


notebooks/omnivoice/omnivoice_helper.py
```diff
@@ -284,7 +284,8 @@ def _patched_encoder_forward(
             for layer in self_.layers:
                 if output_hidden_states:
                     all_hidden = all_hidden + (hidden_states,)
-                hidden_states = layer(hidden_states, attention_mask=None)[0]
+                # transformers>=5 HubertEncoderLayer returns a bare tensor, not a tuple
```
dorloff: Should it be 5.17?

dorloff:
It looks like we need to change the transformers version to ">=5.17" in this notebook.

dorloff (approved):

aleksandr-mokrov (approved):

